PR #1, opened 2017-05-05
Caffe 59

clnowacki:


clnowacki:
Updated the descriptor file as you suggested.

PR #2, opened 2017-05-09
Update AUTO_ROTATE description

brosenberg42:


cpp/OcvFaceDetection/plugin-files/descriptor/descriptor.json
```diff
@@ -171,7 +171,7 @@
     },
     {
       "name": "OCV FACE DETECTION WITH AUTO ORIENTATION ACTION",
```
jrobble: We should be consistent. We have some actions, tasks, pipelines with "WITH AUTO-ROTATION" and we have these ones with "WITH AUTO ORIENTATION" (actually these should be "(WITH AUTO-ORIENTATION)" - note the parentheses).

Now that I think about it, "WITH AUTO-ROTATION" is only half correct. All of the "WITH EXIF" actions, tasks, and pipelines also do AUTO_FLIP. Thus "WITH AUTO-ORIENTATION" is a better choice of words.

Please update all of the components to use "WITH AUTO-ORIENTATION" instead of "WITH AUTO-ROTATION". Also, update these "WITH AUTO ORIENTATION" pipelines to use "(WITH AUTO-ORIENTATION)".

Note that the pipeline name may also need to be updated in the generated JSON output files if the tests don't pass.

PR #15, opened 2017-08-03
Make DlibFace component use frame skipper

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/15)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ install_manifest.txt
 CTestTestfile.cmake
 build/
 cmake-build-debug/
+cmake-build-release/
 
 # Compiled Object files
 *.slo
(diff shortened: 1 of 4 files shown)


jrobble:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/15)*

jrobble:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/15)*

brosenberg42:
> For the most part, I think we can just copy + paste the code for this one. 

Wouldn't it be better to just have the sample binaries call a function in the testUtils library? The only part that is Dlib specific is when the component is created. We could add a function to testUtils that takes the component and command line args as parameters.

jrobble:
> Wouldn't it be better to just have the sample binaries call a function in the testUtils library

Yeah, that would be better. Off the top of my head, the only thing in the component-specific code that might change are the values in the job properties map. The function in the testUtils library should take that map as a parameter.

jrobble:
Additionally, as part of testing feed-forward, we'll probably want a way to construct the job objects with feed-forward track objects. That can be an additional parameter too.

PR #31, opened 2017-10-03
Issue199

clnowacki:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/31)


jrobble:
Did you verify that the example code for calculating the spectral hash produces the same bit string?

---

Reviewed 1 of 4 files at r1.
Review status: 1 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 58 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY3pRgxF15n4uRsMyz:-KvY3pRgxF15n4uRsMz-:b-vo2slt) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L58)):*
> ```cpp
> 
>         if (!name_list.empty()) {
>         boost::trim(name_list);
> ```

Looks like this could be indented.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 118 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYPj0kdy6WcSkV_1hp:-KvYPj0lo64qfuMtJkC_:b-pkq7yh) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L118)):*
> ```cpp
> 
>             }catch(const cv::Exception &err){
>                 LOG4CXX_ERROR(logger_, "Failed to load spectral hash parameter file named "
> ```

Since calculating the spectral hash is optional functionality, like returning activation layer values, when an invalid spectral hash file is provided it should not kill the job. Instead, we should provide an entry in the detectionProperties map. For example:

"FOO SPECTRAL HASH VALUE" = "INVALID"

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 425 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY4yE91wHuKq1pKkIa:-KvY4yE91wHuKq1pKkIb:b-4i137v) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L425)):*
> ```cpp
>     std::vector<cv::Mat> out_mats;
>     net.forward(out_mats, output_layers); // compute output
>     assert(out_mats.size() == output_layers.size());
> ```

What happens if this assert fails? Does the job hang?

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 530 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY5KbDWHTFZ5RP0hxf:-KvY5KbEgYbs1_Ec_LxM:b-e1cmgv) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L530)):*
> ```cpp
>         std::string name = good_names[i];
> 
>         // Flatten the output Mat into a list of floats and convert to
> ```

This looks like a regression. We want to put the JSON in the detectionProperties map.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 586 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYFC00Bfd87tXWx-xQ:-KvYFC01NDUrnHQxBxc3:b-o1muq) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L586)):*
> ```cpp
>     std::string name(hash_data.layer_name);
>     std::transform(name.begin(), name.end(), name.begin(), ::toupper);
>     name += " SPECTRAL_HASH_VALUE";
> ```

Let's use white spaces instead of underscores in this label.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 589 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYFJ6G6TOaAJwzRX3B:-KvYFJ6HEEwr1Y23z2to:ba6xw9b) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L589)):*
> ```cpp
>     LOG4CXX_DEBUG(logger_, "bitset = " << bitset);
>     return(std::make_pair(name, bitset));
>     // return  bits.to_ulong();
> ```

Can this be removed?

---

*[cpp/CaffeDetection/plugin-files/descriptor/descriptor.json, line 202 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY3ZM1x8ViA63C8S2m:-KvY3ZM1x8ViA63C8S2n:b-298mwr) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/plugin-files/descriptor/descriptor.json#L202)):*
> ```JSON
>         },
>         {
>           "name": "SPECTRAL_HASH_FILE_LIST",
> ```

Please move this under ACTIVATION_LAYER_LIST.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-:-KvYQfjUKLELOMVwGc27:b-d6woep)*

jrobble:
Reviewed 3 of 4 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Yes.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 118 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYPj0kdy6WcSkV_1hp:-KvYTnUQmTfw2XznSQFX:bdcv6l8) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L118)):*

Since calculating the spectral hash is optional functionality, like returning activation layer values, when an invalid spectral hash file is provided it should not kill the job. Instead, we should provide an entry in the detectionProperties map. For example:

"FOO SPECTRAL HASH VALUE" = "INVALID"
</blockquote></details>

Good point.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 425 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY4yE91wHuKq1pKkIa:-KvYTqbU3ny32c8rB7S6:b-u43bb3) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L425)):*

What happens if this assert fails? Does the job hang?
</blockquote></details>

Assertions get compiled away unless you define NDEBUG. For a release build, NDEBUG is undefined. To get the assert, you must put "#define NDEBUG" somewhere in the code.

If you would prefer, I could take it out.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 589 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYFJ6G6TOaAJwzRX3B:-KvYUF6RaVdhOX0byfDJ:b2nhswy) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L589)):*

Can this be removed?
</blockquote></details>

Yes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-:-KvYTiz_qKaVPmmriXjH:bb77ufg)*

clnowacki:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 58 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY3pRgxF15n4uRsMyz:-KvYUqAKTI-npkFoeZ-U:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L58)):*

Looks like this could be indented.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 118 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYPj0kdy6WcSkV_1hp:-KvYV8L-VQK0nd50xxWq:b-8p6b7j) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L118)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Good point.
</blockquote></details>

If the FileStorage throws an exception because it can't open the file, then the layer name would not be available for the output line you describe above. I can log a warning to indicate what the filename was that couldn't be opened, but I would have no way of indicating in the output object which layer it was that corresponds to the bad file name. I could put something like "SPECTRAL_HASH_BAD_FILE_NAME" = "bad/path/to/hash/file"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 530 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY5KbDWHTFZ5RP0hxf:-KvYXpJhsUaPs6ncX3Nd:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L530)):*

This looks like a regression. We want to put the JSON in the detectionProperties map.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 586 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYFC00Bfd87tXWx-xQ:-KvYXuAZrV7inSqZlp3f:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L586)):*

Let's use white spaces instead of underscores in this label.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 589 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYFJ6G6TOaAJwzRX3B:-KvYYjDQjM6KZD2cQ_6Q:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L589)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Yes
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/plugin-files/descriptor/descriptor.json, line 202 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY3ZM1x8ViA63C8S2m:-KvYYl9DJTlegQyDQ9ov:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/plugin-files/descriptor/descriptor.json#L202)):*

Please move this under ACTIVATION_LAYER_LIST.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 118 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYPj0kdy6WcSkV_1hp:-KvYZHgExkxyp0o8jLZ8:buz0or6) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L118)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

If the FileStorage throws an exception because it can't open the file, then the layer name would not be available for the output line you describe above. I can log a warning to indicate what the filename was that couldn't be opened, but I would have no way of indicating in the output object which layer it was that corresponds to the bad file name. I could put something like "SPECTRAL_HASH_BAD_FILE_NAME" = "bad/path/to/hash/file"
</blockquote></details>

We'd need a way to list more than one bad special hash file. How about:

"INVALID SPECTRAL HASH FILE LIST" : "<file 1>; <file 2>; ..."

?

If we take this approach, we should update the way we specify bad activation layers:

"INVALID ACTIVATION LAYER LIST" : "<layer 1>; <layer 2>; ..."

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 425 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY4yE91wHuKq1pKkIa:-KvY_6vGohAOFlo28ddV:b1s1q0y) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L425)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Assertions get compiled away unless you define NDEBUG. For a release build, NDEBUG is undefined. To get the assert, you must put "#define NDEBUG" somewhere in the code.

If you would prefer, I could take it out.
</blockquote></details>

If it doesn't cause the job to hang then I'm fine with it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: 1 of 4 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 118 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYPj0kdy6WcSkV_1hp:-KvYftBNBODXOIunGSFt:b2nhll9) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L118)):*

We'd need a way to list more than one bad special hash file. How about:

"INVALID SPECTRAL HASH FILE LIST" : "<file 1>; <file 2>; ..."

?

If we take this approach, we should update the way we specify bad activation layers:

"INVALID ACTIVATION LAYER LIST" : "<layer 1>; <layer 2>; ..."
</blockquote></details>

Ok.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 425 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY4yE91wHuKq1pKkIa:-KvYfmKJAWWICokTV3qP:b2nhll9) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L425)):*

If it doesn't cause the job to hang then I'm fine with it.
</blockquote></details>

Ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: 1 of 4 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 425 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY4yE91wHuKq1pKkIa:-KvbllvsEWaSnDq_ebI9:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L425)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Ok.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 586 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYFC00Bfd87tXWx-xQ:-KvbloMIxRI3RfpVMN8M:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L586)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 589 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvYFJ6G6TOaAJwzRX3B:-KvblrpA8NMS0j9CEHsE:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/CaffeDetection.cpp#L589)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/plugin-files/descriptor/descriptor.json, line 202 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvY3ZM1x8ViA63C8S2m:-Kvblt0gQp661cqBEoHY:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/6da7a3955ea77d6bbd5b31b269b108c13bf16ba9/cpp/CaffeDetection/plugin-files/descriptor/descriptor.json#L202)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Running tests now.

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 82 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbn8d2-bh2X9urGQA4:-Kvbn8d32K-J0O0TsCNd:bgd0zd6) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L82)):*
> ```cpp
>                                     std::vector<std::string> &good_names,
>                                     std::vector<std::string> &bad_layer_names,
>                                     std::vector<std::string> &bad_file_names,
> ```

Let's call this `bad_hash_file_names`, which is used elsewhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-:-Kvbv2SAf_TJ5W5QHTNv:b-ezuxyb)*

jrobble:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-Kvbx4bY3-hXs9BpN-CS:b-25ymlm) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*
> ```cpp
>     MPFDetectionError rc = getSpectralHashInfo(net_layers, hash_file_list,
>                                                good_hash_layer_names,
>                                                bad_hash_layer_names,
> ```

`bad_hash_layer_names` isn't used for anything, but it should be. I ran a test with these params:

```
    // DEBUG
    algorithm_properties["MODEL_NAME"] = "yahoo_nsfw";
    algorithm_properties["NUMBER_OF_CLASSIFICATIONS"] = "2";
    algorithm_properties["RESIZE_HEIGHT"] = "224";
    algorithm_properties["RESIZE_WIDTH"] = "224";
    algorithm_properties["SUBTRACT_BLUE_VALUE"] = "104.0";
    algorithm_properties["SUBTRACT_GREEN_VALUE"] = "117.0";
    algorithm_properties["SUBTRACT_RED_VALUE"] = "123.0";
    // algorithm_properties["ACTIVATION_LAYER_LIST"] = "inception_4c/relu_pool_proj;foo;loss3/classifier;moo;;a";
    algorithm_properties["SPECTRAL_HASH_FILE_LIST"] = "/home/mpf/git/openmpf-projects/openmpf-components/cpp/CaffeDetection/plugin-files/config/bvlc_googlenet_spectral_hash.json";
```

This is what I got:

```
[mpf@localhost build]$ ./sample_caffe_detector /home/mpf/Desktop/TMP/CAFFE/googlenet_0.png 
uri is /home/mpf/Desktop/TMP/CAFFE/googlenet_0.png
detection number 0

CLASSIFICATION is safe for work

CLASSIFICATION CONFIDENCE LIST is 0.998755; 0.00124518

CLASSIFICATION LIST is safe for work; not safe for work
```

Note that there is no mention that the hash file is invalid for the yahoo_nsfw model. 

Note that I modified sample_caffe_detector slightly to run this test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

brosenberg42:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 572 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvcMOlfUHvIxl62CoWs:-KvcMOlfUHvIxl62CoWt:b-w3l0uy) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L572)):*
> ```cpp
>         int bit=1;
>         for(int c=0;c<x.cols;c++){
>             bit *= ((cos(xPtr[x.cols * r + c])>0) ? 1 : -1);
> ```

This pointer stuff is a little bit confusing, I think it would clearer if you changed:
`xPtr[x.cols * r + c]`
to
`x.at<float>(r, c)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-KvdC38TIuwsP9JpSvss:brckroj) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*

`bad_hash_layer_names` isn't used for anything, but it should be. I ran a test with these params:

```
    // DEBUG
    algorithm_properties["MODEL_NAME"] = "yahoo_nsfw";
    algorithm_properties["NUMBER_OF_CLASSIFICATIONS"] = "2";
    algorithm_properties["RESIZE_HEIGHT"] = "224";
    algorithm_properties["RESIZE_WIDTH"] = "224";
    algorithm_properties["SUBTRACT_BLUE_VALUE"] = "104.0";
    algorithm_properties["SUBTRACT_GREEN_VALUE"] = "117.0";
    algorithm_properties["SUBTRACT_RED_VALUE"] = "123.0";
    // algorithm_properties["ACTIVATION_LAYER_LIST"] = "inception_4c/relu_pool_proj;foo;loss3/classifier;moo;;a";
    algorithm_properties["SPECTRAL_HASH_FILE_LIST"] = "/home/mpf/git/openmpf-projects/openmpf-components/cpp/CaffeDetection/plugin-files/config/bvlc_googlenet_spectral_hash.json";
```

This is what I got:

```
[mpf@localhost build]$ ./sample_caffe_detector /home/mpf/Desktop/TMP/CAFFE/googlenet_0.png 
uri is /home/mpf/Desktop/TMP/CAFFE/googlenet_0.png
detection number 0

CLASSIFICATION is safe for work

CLASSIFICATION CONFIDENCE LIST is 0.998755; 0.00124518

CLASSIFICATION LIST is safe for work; not safe for work
```

Note that there is no mention that the hash file is invalid for the yahoo_nsfw model. 

Note that I modified sample_caffe_detector slightly to run this test.
</blockquote></details>

Fixed that.
I guess we should include the model name in the spectral hash input file so that we can double check that too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 82 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbn8d2-bh2X9urGQA4:-KvdCM7LszQEveQpTGdJ:bb6un1j) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L82)):*

Let's call this `bad_hash_file_names`, which is used elsewhere.
</blockquote></details>

Done

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 572 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvcMOlfUHvIxl62CoWs:-KvdCIwZNDOjfGM-N5xf:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L572)):*

This pointer stuff is a little bit confusing, I think it would clearer if you changed:
`xPtr[x.cols * r + c]`
to
`x.at<float>(r, c)`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-KveE0NU5zO7x7eED3qg:bcqvh3q) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*
> I guess we should include the model name in the spectral hash input file so that we can double check that too.

I can imagine a case where two models are trained using the same kind of network (same layers and number of nodes), so it's possible to use the same spectral hash JSON for two separate models.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-KvggWn2CqNuKaL1idrV:b-nclvt0) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*

> I guess we should include the model name in the spectral hash input file so that we can double check that too.

I can imagine a case where two models are trained using the same kind of network (same layers and number of nodes), so it's possible to use the same spectral hash JSON for two separate models.
</blockquote></details>

So I'm not sure, then, how we could detect that a hash file is invalid for a particular model, except to point out that the layer name was invalid.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-KvgorGG_bDx5utCIyPl:b-hb4sik) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

So I'm not sure, then, how we could detect that a hash file is invalid for a particular model, except to point out that the layer name was invalid.
</blockquote></details>

That's what I was thinking. If the layer name is not in the model then we add the file to the list of invalid hash files. We should log an appropriate message to the log file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-KvgqfWFjYuS31fyDFmq:b4pupih) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*

That's what I was thinking. If the layer name is not in the model then we add the file to the list of invalid hash files. We should log an appropriate message to the log file.
</blockquote></details>

So if I understand what you are saying, there would only ever be an invalid hash file list, not an invalid layer name list. For the activation values, we report invalid layers, but from what you are saying, invalid spectral hash layers would go into the invalid file list, so no need for an invalid layers list?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-Kvgrq0D2EWNKlNpNpB9:bkjs2l9) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

So if I understand what you are saying, there would only ever be an invalid hash file list, not an invalid layer name list. For the activation values, we report invalid layers, but from what you are saying, invalid spectral hash layers would go into the invalid file list, so no need for an invalid layers list?
</blockquote></details>

I think you will always need an "INVALID ACTIVATION LAYER LIST" to represent invalid "ACTIVATION_LAYER_LIST" entries.

If a file listed in the "SPECTRAL_HASH_FILE_LIST" does not exist, cannot be parsed, or has an invalid layer in it for the specified model, then we add the hash file to the "INVALID SPECTRAL HASH FILENAME LIST".

The user will have to look at the log messages to determine which of those three error conditions occurred.

How's that sound? Do you have a different idea?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-KvguGaYCenVsWcSC_Xn:bk4qoan) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*

I think you will always need an "INVALID ACTIVATION LAYER LIST" to represent invalid "ACTIVATION_LAYER_LIST" entries.

If a file listed in the "SPECTRAL_HASH_FILE_LIST" does not exist, cannot be parsed, or has an invalid layer in it for the specified model, then we add the hash file to the "INVALID SPECTRAL HASH FILENAME LIST".

The user will have to look at the log messages to determine which of those three error conditions occurred.

How's that sound? Do you have a different idea?
</blockquote></details>

Yes, that's exactly what I was thinking.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-Kvgv2tRfju74v_B1Wfe:b-5f677s) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Yes, that's exactly what I was thinking.
</blockquote></details>

OK, sounds good. Please do it that way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Reviewed 1 of 3 files at r2.
Review status: 2 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 410 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvbx4bXggh5lG6OmLx8:-KvlnO0zvtf1qpzHS6Yd:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L410)):*

OK, sounds good. Please do it that way.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Review status: 2 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 81 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm6v-KnVw8TdckjDF2:-Kvm6v-KnVw8TdckjDF3:b-1e3na2) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L81)):*
> ```cpp
>                                     const std::string &model_name,
>                                     std::string &hash_file_list,
>                                     std::vector<std::string> &good_names,
> ```

This should be `good_hash_layer_names`.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 108 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmBvDwMRDmbumoZ6BH:-KvmBvDwMRDmbumoZ6BI:bonh1kg) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L108)):*
Should we be doing something with the `nbits` parameter? Minimally, maybe we should be using it for verification somehow.

I removed "mx" from the spectral hash file and got:

> OpenCV Error: Sizes of input arguments do not match (The operation is neither 'array op array' (where arrays have the same size and the same number of channels), nor 'array op scalar', nor 'scalar op array') in arithm_op, file /apps/source/opencv_sources/opencv/modules/core/src/arithm.cpp, line 659

I think it would be better if we verified that the JSON contained each of the values you're expecting to see in `spParams`. If one is missing, print out a warning and mark the hash file as bad. 

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 124 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmAHwJJMWdoG4qKS4a:-KvmAHwJJMWdoG4qKS4b:b-lf2eoh) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L124)):*
> ```cpp
>             } catch(const cv::Exception &err){
>                 LOG4CXX_WARN(logger_, "Exception caught when processing spectral hash file named \"" << filename << "\": " << err.what());
>                 bad_hash_file_names.push_back(filename);
> ```

Whenever the spectral hash file is pushed onto the collection of bad hash files, have the warning message also state the name of the file and that it's invalid.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 421 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm7ItKWg-2vf8OuX-_:-Kvm7ItLF5BRC-21KVJ2:b93n6ni) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L421)):*
> ```cpp
>     }
> 
>     output_layers.push_back(cv::String(output_layer_name));
> ```

What happens if the output layer is also specified as an activation layer or spectral hash layer? Or an activation layer is also specified as a spectral hash layer? Could `output_layers` contain duplication entries? I think the answer is yet. That means `output_layers` is larger than it needs to be.

Instead, should we be using a map to manage the output layers so we can access them by name?

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 562 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmBhjWcZjG0YYlR3ZA:-KvmBhjXoEJOttY8Mc8B:b46wuks) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L562)):*
> ```cpp
> std::pair<std::string,std::string>
> CaffeDetection::computeSpectralHash(const cv::Mat &activations,
>                                     const SpectralHashInfo &hash_data) {
> ```

Here you call it `hash_data`, but elsewhere you call it `hash_info`.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 101 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm6ql0IHWiyphMLdRE:-Kvm6ql0IHWiyphMLdRF:b-1e3na2) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.h#L101)):*
> ```objc
>                         const std::string &model_name,
>                         std::string &hash_file_list,
>                         std::vector<std::string> &good_names,
> ```

This should be `good_hash_layer_names`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 70 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmE4OoEAC_1SU0gGLb:-KvmE4Opfr7NjXAFdGth:bch78ax) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L70)):*
> ```cpp
>                 }
>                 else {
>                     bad_names.push_back(name);
> ```

Print out a warning here like:

`LOG4CXX_WARN(logger_, "Layer named \"" << name << "\" was not found in the model named \"" << model_name << "\"");`

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 421 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm7ItKWg-2vf8OuX-_:-KvmDhhDmBtcIkMHiD9q:b-7ucfo4) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L421)):*

What happens if the output layer is also specified as an activation layer or spectral hash layer? Or an activation layer is also specified as a spectral hash layer? Could `output_layers` contain duplication entries? I think the answer is yet. That means `output_layers` is larger than it needs to be.

Instead, should we be using a map to manage the output layers so we can access them by name?
</blockquote></details>

* is yes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 124 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmAHwJJMWdoG4qKS4a:-KvmIdPXsN387Iknm9Sm:b-mj7yif) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L124)):*

Whenever the spectral hash file is pushed onto the collection of bad hash files, have the warning message also state the name of the file and that it's invalid.
</blockquote></details>

The message does contain the name of the file. Do you need more than the fact that an exception was caught to know that the file was invalid??

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 421 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm7ItKWg-2vf8OuX-_:-KvmIqxn1VEZ_ZBrEqIO:b-lfousm) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L421)):*

* is yes
</blockquote></details>

If the output layer is also specified as an activation layer, you get both outputs. If an activation layer is also specified as a spectral hash layer, you get both as well. I've tested that. The internal OpenCV code does not check for duplicate layer names.

The net.forward function requires a vector of strings as input, and returns a vector of mats. I think it would be unnecessarily complicated to be moving things back and forth between vectors and a map.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 124 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmAHwJJMWdoG4qKS4a:-KvmLm3mkjKHkg-OfnhE:b-siodjx) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L124)):*
> LOG4CXX_WARN(logger_, "Layer named \"" << hash_info.layer_name << "\" was not found in the model named \"" << model_name << "\"");

This is the message I saw in the log. In retrospect, this is the only message that seems like it needs to be updated with the hash file name.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 421 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm7ItKWg-2vf8OuX-_:-KvmM2RZUQ2wdv6w_atZ:b-lk3mv) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L421)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

If the output layer is also specified as an activation layer, you get both outputs. If an activation layer is also specified as a spectral hash layer, you get both as well. I've tested that. The internal OpenCV code does not check for duplicate layer names.

The net.forward function requires a vector of strings as input, and returns a vector of mats. I think it would be unnecessarily complicated to be moving things back and forth between vectors and a map.
</blockquote></details>

It seems like these Mats can be quite large and we might be eating up more memory than we need. I assume that pointers are used under the hood so we're not duplicating memory usage, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 124 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmAHwJJMWdoG4qKS4a:-KvmO3wyEcWn4oBWhH1R:b-4gor3b) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L124)):*

> LOG4CXX_WARN(logger_, "Layer named \"" << hash_info.layer_name << "\" was not found in the model named \"" << model_name << "\"");

This is the message I saw in the log. In retrospect, this is the only message that seems like it needs to be updated with the hash file name.
</blockquote></details>

Gotcha.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 421 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm7ItKWg-2vf8OuX-_:-KvmNQn__rcX_zzbeXyW:b-ck9o8s) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L421)):*

It seems like these Mats can be quite large and we might be eating up more memory than we need. I assume that pointers are used under the hood so we're not duplicating memory usage, right?
</blockquote></details>

They are cv::Mat objects, so all they are is headers which are quite small. The data is in one place and the header simply points to it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
*[cpp/CaffeDetection/CaffeDetection.cpp, line 421 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm7ItKWg-2vf8OuX-_:-KvmOaTnOB-ZW8ktWfk8:b-v5kin0) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L421)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

They are cv::Mat objects, so all they are is headers which are quite small. The data is in one place and the header simply points to it.
</blockquote></details>

Ok, then no need to change anything here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: 2 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 70 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmE4OoEAC_1SU0gGLb:-Kw0Wky-PiH6hCpXmEHG:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L70)):*

Print out a warning here like:

`LOG4CXX_WARN(logger_, "Layer named \"" << name << "\" was not found in the model named \"" << model_name << "\"");`
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 81 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm6v-KnVw8TdckjDF2:-KvmOYvvZhfRhwB4YlN6:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L81)):*

This should be `good_hash_layer_names`.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 108 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmBvDwMRDmbumoZ6BH:-Kw0Wmb6oqYe485MuFoe:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L108)):*

Should we be doing something with the `nbits` parameter? Minimally, maybe we should be using it for verification somehow.

I removed "mx" from the spectral hash file and got:

> OpenCV Error: Sizes of input arguments do not match (The operation is neither 'array op array' (where arrays have the same size and the same number of channels), nor 'array op scalar', nor 'scalar op array') in arithm_op, file /apps/source/opencv_sources/opencv/modules/core/src/arithm.cpp, line 659

I think it would be better if we verified that the JSON contained each of the values you're expecting to see in `spParams`. If one is missing, print out a warning and mark the hash file as bad. 

</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 562 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmBhjWcZjG0YYlR3ZA:-Kw0Wi-sEaoOW7FwhWqj:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L562)):*

Here you call it `hash_data`, but elsewhere you call it `hash_info`.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 101 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kvm6ql0IHWiyphMLdRE:-Kw0WjrmetFjmd2LcvBx:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.h#L101)):*

This should be `good_hash_layer_names`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
*[cpp/CaffeDetection/CaffeDetection.cpp, line 562 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmBhjWcZjG0YYlR3ZA:-Kw1CrdR6cxzcVIXEgYf:bbujnfg) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L562)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Done.
</blockquote></details>

You're still using both in your most recent code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: 2 of 4 files reviewed at latest revision, 8 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 82 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw1Am5B91Z6oRoBD5vc:-Kw5UV8kANuIHfR4YeDj:b-cak53w) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L82)):*

In some places you use `hash_info` and other places you use `hashInfo_`. I think the former is more consistent with the rest of your style. Please use that everywhere.
</blockquote></details>

The variable hashInfo_ is a class member.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 92 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw17zth25pDYgDjgN4w:-Kw5h_TKy56ZTutw_zEp:b-rhwuh6) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L92)):*

Please verify that `nbits` exists first.
</blockquote></details>

I'm not sure why you think we need to do this. I've tested running this with a modified hash file, where both the nbits node and the mx node have been deleted, and the appropriate warnings appear in the log file and the file is flagged as a bad file.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 143 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw18QHKSuce1z6p6PrS:-Kw5mc9LSX8GAF2j6HHN:b-otilcs) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L143)):*

Please verify that `layer_name` exists before using it.
</blockquote></details>

See comment above.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 86 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw1ASTJH34XIsUAurGz:-Kw5URS6bX2-PI7I3wxX:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.h#L86)):*

The indenting of some functions is off.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: 2 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 562 at r3](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvmBhjWcZjG0YYlR3ZA:-Kw5sRzAlbM9NUvqqKjo:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/999829df4faf80c84bd4ceb8a3e34bc258b91ba0/cpp/CaffeDetection/CaffeDetection.cpp#L562)):*

You're still using both in your most recent code.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 162 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw1CK4nrcylfb6Kuyri:-Kw5s6saleRGFQoJgsLR:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L162)):*

I would rephrase this to:

"Layer named XXX from spectral hash file YYY was not found in the model named ZZZ"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 82 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw1Am5B91Z6oRoBD5vc:-Kw65Rf1rf_Moz5cK61A:bqsnmv9) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L82)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

The variable hashInfo_ is a class member.
</blockquote></details>

It's used as a parameter here:

```
MPFDetectionError
CaffeDetection::getSpectralHashInfo(const std::vector<cv::String> &names_to_search,
                                    const std::string &model_name,
                                    std::string &hash_file_list,
                                    std::vector<std::string> &good_hash_layer_names,
                                    std::vector<std::string> &bad_hash_file_names,
                                    std::vector<SpectralHashInfo> &hashInfo_) {
```

It is necessary to pass it as a parameter there?

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 92 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw17zth25pDYgDjgN4w:-Kw64ZUHEt0MRXK5CbLd:b-gsl0az) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L92)):*
When nbits was not specified I got the warning:

> The "nbits" field in file FFF is less than or equal to zero.

That's not accurate. The "nbits" field didn't exist.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 143 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw18QHKSuce1z6p6PrS:-Kw64uUI-PzhaoVIlNu2:b-z5ncao) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L143)):*
When I removed "layer_name" I got the following warning:

> Layer name "" was not found ...

Which is a bit odd since "layer_name" didn't exist in the file. That's different from it being set to the value of a blank string.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 82 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw1Am5B91Z6oRoBD5vc:-Kw67lnakqdNDKcfXgH6:b-psiiza) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L82)):*

It's used as a parameter here:

```
MPFDetectionError
CaffeDetection::getSpectralHashInfo(const std::vector<cv::String> &names_to_search,
                                    const std::string &model_name,
                                    std::string &hash_file_list,
                                    std::vector<std::string> &good_hash_layer_names,
                                    std::vector<std::string> &bad_hash_file_names,
                                    std::vector<SpectralHashInfo> &hashInfo_) {
```

It is necessary to pass it as a parameter there?

</blockquote></details>

Nope. not necessary at all, you're right

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 92 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw17zth25pDYgDjgN4w:-Kw67qJHfdaIHBHWFZB7:bodwqre) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L92)):*

When nbits was not specified I got the warning:

> The "nbits" field in file FFF is less than or equal to zero.

That's not accurate. The "nbits" field didn't exist.
</blockquote></details>

I get your point now. Thanks

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 143 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw18QHKSuce1z6p6PrS:-Kw67x5AdH7aoIJEj_FR:b-4gor3b) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L143)):*

When I removed "layer_name" I got the following warning:

> Layer name "" was not found ...

Which is a bit odd since "layer_name" didn't exist in the file. That's different from it being set to the value of a blank string.

</blockquote></details>

Gotcha.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 82 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw1Am5B91Z6oRoBD5vc:-Kw6wWz-cWbDtR4J5crU:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L82)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Nope. not necessary at all, you're right
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 92 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw17zth25pDYgDjgN4w:-Kw70M_AlonIazzPRAan:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L92)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

I get your point now. Thanks
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 143 at r4](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-Kw18QHKSuce1z6p6PrS:-Kw70OM3Czf8fULsOcOF:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/38f544bd95b62f5efc926a86d85cf7ebd21ad726/cpp/CaffeDetection/CaffeDetection.cpp#L143)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Gotcha.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

jrobble:
@clnowacki, are you going to make an update to this to validate the `nbits` parameter?

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-:-KwB_sY1n_mVW2RANf2T:b-s9lzwb)*

clnowacki:
Yes, coming right up.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-:-KwBmaFLlNrk2ywbM_Bp:bdpo7vq)*

clnowacki:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 572 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-KvcMOlfUHvIxl62CoWs:-KwBp9QJssAap2Q8Xg_c:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/5fee9e7e12d92d5a10058f572468973d785b926e/cpp/CaffeDetection/CaffeDetection.cpp#L572)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

clnowacki:
Reviewed 1 of 4 files at r1, 1 of 3 files at r2, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

brosenberg42:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31#-:-KwBrH_3_WwyJMZzL9dT:bnfp4nl)*

jrobble:
Reviewed 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/31)*

PR #35, opened 2017-10-11
Add video support to Caffe.

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/35)


jrobble:
Please add the following to descriptor.json:

```
    "providesCollection": {
      "states": [
        "DETECTION",
        "DETECTION_CLASS",
        "DETECTION_CLASS_CAFFE"
      ],
```

That will enable users to create a Caffe --> Markup pipeline.

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-:-KwGicB_X6EWcQB5c-aC:b-ofcstz)*

jrobble:
As we discussed, please add support for FEED_FORWARD_TYPE = "REGION".

---

Reviewed 8 of 10 files at r1.
Review status: 8 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/CaffeDetection/test/test_caffe_detection.cpp, line 132 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwMQUHAK72gcOH_q9Rr:-KwMQUHBlEBpkTOpcvGD:b-lup8f1) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/test/test_caffe_detection.cpp#L132)):*
> ```cpp
> 
> 
> TEST(CAFFE, SpectralHashTest) {
> ```

Thanks for adding a test for this even though you didn't write the spectral hash code :thumbsup:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-:-KwMSbWbnHtvLrWRcSf2:b3fsash)*

jrobble:
Reviewed 2 of 10 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 135 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwMUgn0MgK1ivE134TU:-KwMUgn13fslV8PzVi1o:bf1k6dq) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.cpp#L135)):*
> ```cpp
> void defaultTracker(MPFImageLocation &location, int frame_index, std::vector<MPFVideoTrack> &tracks) {
>     bool should_start_new_track = tracks.empty()
>                                || tracks.back().detection_properties["CLASSIFICATION"]
> ```

As you pointed out, please only check the highest confidence classification, which should be first in the semicolon-delimited list.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 196 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwMW8BuX_h1thvM7meI:-KwMW8Bv2wgO0iUxYm5E:b-tvxkmt) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.cpp#L196)):*
> ```cpp
>             }
>             if (!location) {
>                 // Nothing found in current frame.
> ```

Note to self: If no classifications are above the confidence threshold then `location` could be null.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 124 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwMa4tjEC5IHNDvWHv9:-KwMa4tjEC5IHNDvWHvA:b1coqlp) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.h#L124)):*
> ```objc
> 
>     // struct to hold configuration options and data structures that change every job.
>     struct JobConfig {
> ```

This struct makes a lot of sense. To avoid confusion with MPFJob objects, let's call this `CaffeJobConfig`.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 157 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwM_Q7e5vsVrfTfMyby:-KwM_Q7fHuSjGDuFFqgu:byp0ir0) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.h#L157)):*
> ```objc
> 
>         JobConfig(const MPF::COMPONENT::Properties &props, const std::map<std::string, ModelFiles> &model_defs,
>                   const log4cxx::LoggerPtr &logger);
> ```

If the logger is passed in the constructor, does it need to be passed in the three functions below? Can it be a data member of this struct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/35)*

brosenberg42:
> As we discussed, please add support for FEED_FORWARD_TYPE = "REGION".

Those changes will be done in openmpf-cpp-sdk, so it will be done in a separate pull request.

---

Review status: 8 of 10 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 135 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwMUgn0MgK1ivE134TU:-Kw_K25jxys8t5cfmJLJ:b1t5lo6) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.cpp#L135)):*

As you pointed out, please only check the highest confidence classification, which should be first in the semicolon-delimited list.
</blockquote></details>

Originally this used the "CLASSIFICATION LIST" property and was using the entire list for tracking, now it uses the "CLASSIFICATION" property so that it only uses the highest confidence classification.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 124 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwMa4tjEC5IHNDvWHv9:-Kw_Kkz8RSb4VNZiHDGi:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.h#L124)):*

This struct makes a lot of sense. To avoid confusion with MPFJob objects, let's call this `CaffeJobConfig`.
</blockquote></details>

Done.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 157 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwM_Q7e5vsVrfTfMyby:-Kw_KmEk55SrDIPEJRU6:b-tcjfhd) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.h#L157)):*

If the logger is passed in the constructor, does it need to be passed in the three functions below? Can it be a data member of this struct?
</blockquote></details>

I would prefer not to. The logger is only used during initialization, after that it is no longer used. Those methods are only called by the constructor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-:-Kw_Jr9MR3THPk8GmyGM:budli8e)*

jrobble:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[cpp/CaffeDetection/CaffeDetection.cpp, line 135 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwMUgn0MgK1ivE134TU:-Kw_PVEd7jxh4SFVLc1d:bdcv6l8) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.cpp#L135)):*

Originally this used the "CLASSIFICATION LIST" property and was using the entire list for tracking, now it uses the "CLASSIFICATION" property so that it only uses the highest confidence classification.
</blockquote></details>

Good point.

---

*[cpp/CaffeDetection/CaffeDetection.h, line 157 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/35#-KwM_Q7e5vsVrfTfMyby:-Kw_PaidLTcNJrh3GLOS:bbz38od) ([raw file](https://github.com/openmpf/openmpf-components/blob/1c988d8dcc07ef88dd9d53a8ff13ca8c6ed70747/cpp/CaffeDetection/CaffeDetection.h#L157)):*

I would prefer not to. The logger is only used during initialization, after that it is no longer used. Those methods are only called by the constructor.
</blockquote></details>

Ok, I'm fine with that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/35)*

PR #52, opened 2018-03-08
Add Darknet component

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/52)


jrobble:
Which files in darknet_lib did you modify, if any?

---

Your component needs the following files in the top-level directory:

- COPYING
- NOTICE
- LICENSE
- README.md

Refer to openmpf-contrib-components/cpp/motion/LICENSE and the way that it mentions STRUCK and Eigen code. The Darknet component license should mention the Darknet code in a similar way.

---

plugin-files/config/cfg contains a lot of files that we don't seem to be using. Can we remove everything but what we need? Similarly, do we need everything in /plugin-files/config/cfg/data?

---

Running "TINY YOLO OBJECT DETECTION (WITH MARKUP) PIPELINE" makes me wish that our markup had the classification labels, much like the YOLO vid on the author's website. I'll consider that for a wishlist task.

I ended up creating a "TINY YOLO OBJECT DETECTION PREPROCESSOR (WITH MARKUP) PIPELINE" for testing.

---

Reviewed 6 of 166 files at r1.
Review status: 6 of 166 files reviewed at latest revision, 9 unresolved discussions.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 107 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aXShwkHOvHm367YMJ:-L7aXShwkHOvHm367YMK:b-w2suus) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L107)):*
> ```cpp
> 
> 
> MPFDetectionError DarknetDetection::GetDetections(const MPFImageJob &job, std::vector<MPFImageLocation> &locations) {
> ```

USE_PREPROCESSOR should work on images too. That should result in using the superset regions, just like it does for videos.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 115 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aS_u9VyyfWYfgR-eZ:-L7aS_u9VyyfWYfgR-e_:bajarlx) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L115)):*
> ```cpp
>         int number_of_classifications
>                 = std::max(1, DetectionComponentUtils::GetProperty(job.job_properties,
>                                                                    "NUMBER_OF_CLASSIFICATIONS_PER_REGION", 5));
> ```

Instead of hardcoding the default value, the convention is to use an .ini file.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 139 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aVBwSgMTjATW3-neU:-L7aVBwSgMTjATW3-neV:br0rd3c) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L139)):*
> ```cpp
> 
> std::string DarknetDetection::GetDetectionType() {
>     return "OBJECT";
> ```

I think this should be "CLASS" to match Caffe. I understand that this also returns the detection region, so it's different, but it is performing classification.

---

*[cpp/DarknetDetection/Trackers.cpp, line 94 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aKAA8zj_IIg4K-Mxx:-L7aKAA8zj_IIg4K-Mxy:b-tcpsrf) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L94)):*
> ```cpp
>     const auto &rect = detection.detection_rect;
> 
>     return MPFImageLocation(rect.x, rect.y, rect.width, rect.height, top_confidence, {
> ```

The Caffe component generates the following:

```
"detectionProperties": {
    "CLASSIFICATION": "golden retriever",
    "CLASSIFICATION CONFIDENCE LIST": "0.984752",
    "CLASSIFICATION LIST": "golden retriever"
},
```

For consistency, I think the Darknet component should use the same keys.

---

*[cpp/DarknetDetection/Trackers.cpp, line 173 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aT_L1HEASu07EIs1J:-L7aT_L1HEASu07EIs1K:bzgrg4r) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L173)):*
> ```cpp
>     frame_location.confidence = frame_location.confidence + prob - frame_location.confidence * prob;
> 
>     track.confidence = std::max(track.confidence, frame_location.confidence);
> ```

FYI, track confidence currently isn't used for anything. See [this issue](https://github.com/openmpf/openmpf/issues/319). Note that specifying it is not a problem; it's just that nothing will be done with it. It won't show up in the JSON output.

---

*[cpp/DarknetDetection/plugin-files/descriptor/descriptor.json, line 26 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aMwpWJ1BevI3fjCki:-L7aMwpXYih89Xk92egg:b-pfoi8j) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/plugin-files/descriptor/descriptor.json#L26)):*
> ```JSON
>         "DETECTION_OBJECT_DARKNET"
>       ],
>       "properties": [
> ```

I see the following unique properties:

- NETWORK_CONFIG_FILE
- WEIGHTS_FILE
- NAMES_FILE
- USE_PREPROCESSOR
- NUMBER_OF_CLASSIFICATIONS_PER_REGION

The convention is to list the unique properties after the list of common properties. That makes it easier to compare the descriptor files for consistency.

Instead of requiring the user to specify three properties (NETWORK_CONFIG_FILE, WEIGHTS_FILE, NAMES_FILE), let's consider creating a .json file that serves the same purpose as openmpf-components/cpp/CaffeDetection/plugin-files/config/models.cfg. That file groups the models and their associated files together so then all the user needs to do is specify the value of MODEL_NAME.

---

*[cpp/DarknetDetection/plugin-files/descriptor/descriptor.json, line 53 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aOYWahBYnzVzZzWg5:-L7aOYWahBYnzVzZzWg6:b-8ch9q7) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/plugin-files/descriptor/descriptor.json#L53)):*
Add:

> The value must be greater than 0 and less than or equal to the number of classifications that the model supports.

---

*[cpp/DarknetDetection/test/CMakeLists.txt, line 1 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aP691jlV9ug6O6S9E:-L7aP6928z4wxXV81guJ:b-ja49be) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/CMakeLists.txt#L1)):*
> ```CMake
> cmake_minimum_required(VERSION 3.6)
> ```

Missing header comment block.

---

*[cpp/DarknetDetection/test/data/NOTICE, line 5 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aQ4Maar0JvXdqEPIA:-L7aQ4MbmLqKdA6WjxnT:b-ncuoc3) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/data/NOTICE#L5)):*
> ```
> Creative Commons Attribution 2.0 Generic License: https://creativecommons.org/licenses/by/2.0/
> https://www.flickr.com/photos/zombieite/12041174763/in/photolist-jm3aKr-hPa71A-e2FUq5-fyKUoT-7FSnzE-fgixCd-bcFXqT-5GUagk-8YA7hh-5V5AwV-7Deq1Z-gcwnJu-8jdjQG-7Kh3hj-86vQ2E-4EdgLD-6sNTxr-f2gXTy-burH8d-8CdB8V-7fzvt8-gdxefs-8ahfVa-g4qGZT-gjMjTZ-79trhD-7dpLRD-86wTHa-99USLo-dcjipF-8SknPC-aE5gM7-e75WMA-6H5bp2-7dPdd3-59p1g1-9Cs2ef-eK5Ws1-6MYmzo-8S5NAT-ij1ZjZ-8niwck-fL81fV-dfEXV7-frNHMq-bsSzvH-ehmDHp-7pQbY6-6Aq9LL-9oHyVj
> First 500ms of original
> ```

For completeness, also list where dog.jpg came from.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-:-L7aLIfHeWgLl41EL8aV:b-i8ccrh)*

jrobble:
When the tests are run on Jenkins I see this kind of output:

```
[ RUN      ] Darknet.ImageTest
layer     filters    size              input                output
    0 conv     16  3 x 3 / 1   416 x 416 x   3   ->   416 x 416 x  16
    1 max          2 x 2 / 2   416 x 416 x  16   ->   208 x 208 x  16
    2 conv     32  3 x 3 / 1   208 x 208 x  16   ->   208 x 208 x  32
    3 max          2 x 2 / 2   208 x 208 x  32   ->   104 x 104 x  32
    4 conv     64  3 x 3 / 1   104 x 104 x  32   ->   104 x 104 x  64
    5 max          2 x 2 / 2   104 x 104 x  64   ->    52 x  52 x  64
    6 conv    128  3 x 3 / 1    52 x  52 x  64   ->    52 x  52 x 128
    7 max          2 x 2 / 2    52 x  52 x 128   ->    26 x  26 x 128
    8 conv    256  3 x 3 / 1    26 x  26 x 128   ->    26 x  26 x 256
    9 max          2 x 2 / 2    26 x  26 x 256   ->    13 x  13 x 256
   10 conv    512  3 x 3 / 1    13 x  13 x 256   ->    13 x  13 x 512
   11 max          2 x 2 / 1    13 x  13 x 512   ->    13 x  13 x 512
   12 conv   1024  3 x 3 / 1    13 x  13 x 512   ->    13 x  13 x1024
   13 conv    512  3 x 3 / 1    13 x  13 x1024   ->    13 x  13 x 512
   14 conv    425  1 x 1 / 1    13 x  13 x 512   ->    13 x  13 x 425
   15 detection
mask_scale: Using default '1.000000'
Loading weights from ../plugin/DarknetDetection/config/weights/tiny-yolo.weights...Done!
[       OK ] Darknet.ImageTest (3145 ms)
```

Is it possible to squelch printing out the network? It's not a big deal, but it is clutter.

---

Reviewed 55 of 166 files at r1.
Review status: 61 of 166 files reviewed at latest revision, 21 unresolved discussions.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 179 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7farUehTUuqlFDAu4U:-L7farUehTUuqlFDAu4V:bpyn6ur) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L179)):*
> ```cpp
> 
> DarknetDetection::ProbHolder::ProbHolder(int output_layer_size, int num_classes)
>     : mat_size_(output_layer_size * (num_classes + 1))
> ```

Why the +1 on `num_classes + 1`? Looking at this code:


```
      for (int name_idx = 0; name_idx < num_classes_; name_idx++) {
            float prob = probs_[box_idx][name_idx];
            if (prob >= confidence_threshold_) {
                detection.object_type_probs.emplace_back(prob, names_[name_idx]);
            }
        }
```

If there are 80 classes, won't `name_idx` be 0 ... 79?

I assume it has something to do with how `probs` is populated here:

```
    network_detect(network_.get(), image.dnet_image, confidence_threshold_, 0.5, 0.3, boxes_.get(),
                   probs_.Get());
```

A comment would be helpful.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 212 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fdZY3bDB_8GpIJcb2:-L7fdZY3bDB_8GpIJcb3:beeg65w) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L212)):*
> ```cpp
>     , num_classes_(GetNumClasses(*network_))
>     , names_(LoadNames(props, num_classes_))
>     , confidence_threshold_(std::max(0.01f, DetectionComponentUtils::GetProperty(props, "CONFIDENCE_THRESHOLD", 0.01f)))
> ```

Why must you set 0.01f as the minimum? What would happen if you set 0?

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 224 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fd-tgHfBPVDwb7nmS:-L7fd-tiiWU2oGJekJBL:b6tjp9v) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L224)):*
> ```cpp
>     DNetImageHolder image(cv_image);
>     probs_.Clear();
>     network_detect(network_.get(), image.dnet_image, confidence_threshold_, 0.5, 0.3, boxes_.get(),
> ```

A comment explaining why we chose 0.5 and 0.3 would be helpful here.

---

*[cpp/DarknetDetection/DarknetDetection.h, line 137 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fJKqEDZUgEA9MfbBg:-L7fJKqEDZUgEA9MfbBh:br6da2c) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.h#L137)):*
> ```objc
>     // One is that it adds a destructor, which calls the free_image function defined in the Darknet library.
>     // The second feature is that it converts a cv::Mat to the Darknet image format.
>     struct DNetImageHolder {
> ```

Let's call this "DarknetImageHolder". We don't shorten darknet anywhere else.

---

*[cpp/DarknetDetection/Trackers.cpp, line 83 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7eh0wrTZW2x5HfS_jr:-L7eh0wrTZW2x5HfS_js:b-egmr9q) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L83)):*
> ```cpp
>     ++prob_pair_iter;
> 
>     std::ostringstream other_confidences;
> ```

The terms `other_confidences` and `other_types` is a little confusing. What is the "other" in this case? Those lists also include the top confidence and type.

---

*[cpp/DarknetDetection/Trackers.cpp, line 102 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7ehZ4_xQ_NS31KYH5Y:-L7ehZ4_xQ_NS31KYH5Z:b-po7kfz) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L102)):*
> ```cpp
> 
> std::vector<MPF::COMPONENT::MPFVideoTrack> SingleDetectionPerTrackTracker::GetTracks() {
>     return std::move(tracks_);
> ```

I'm looking at what we did at the end of `SubsenseStreamingDetection::EndSegment()`:

```
    vector<MPFVideoTrack> tmp;
    tmp.swap(tracks_);
    return tmp;
```

We had a [long conversation](https://reviewable.io/reviews/openmpf/openmpf-contrib-components/36#-L5AOkbEsckaset0hrwx) about using `move` and leaving things in an unspecified state.

I think you're assuming that `GetTracks()` is called just before this class is destructed. Unlike streaming, where `EndSegment()`can be called multiple times during the component's lifetime, `GetTracks()` should only be called once by the component executor. It might be worth mentioning that in a comment in the `GetTracks()` method for both of the tracker types.

---

*[cpp/DarknetDetection/Trackers.cpp, line 113 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fXXGwkCISiBse2UOQ:-L7fXXGyrZlDacCpA28Y:bnl88pk) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L113)):*
> ```cpp
>             auto it = tracks_.find({frame_number - 1, class_prob.second});
>             if (it != tracks_.cend()) {
>                 AddNewImageLocationToTrack(location.detection_rect, class_prob.first, class_prob.second, frame_number,
> ```

Shouldn't we try to combine image locations before adding a new image location to the track?

Because of the other you have these, it seems that if there exists a track that ends on the previous frame which has the same object type as our new detection, then that will always result in adding a new image location to the track.

Actually ... adding a new image location changes the key in the tracks map, so the next new detection in the loop will get combined. Never mind. It may be worth adding a comment here explaining this. 

---

*[cpp/DarknetDetection/Trackers.cpp, line 182 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7em6ovqd4KZk1TmOEE:-L7em6oxiStm9YgpC6Sp:bulk7iq) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L182)):*
> ```cpp
> 
>     for (auto &pair : tracks_) {
>         results.push_back(std::move(pair.second));
> ```

This `move` seems like it will leave the original tracks collection in a partially unspecified state.

---

*[cpp/DarknetDetection/darknet_lib/src/classifier.h, line 1 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fenDScZmFdL45K_19:-L7fenDScZmFdL45K_1A:b24t6e0) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/darknet_lib/src/classifier.h#L1)):*
> ```objc
> 
> ```

Odd that this is empty.

---

*[cpp/DarknetDetection/test/test_darknet_detection.cpp, line 283 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7e_vNhKqGhXykWyCHA:-L7e_vNhKqGhXykWyCHB:b3h0o1i) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/test_darknet_detection.cpp#L283)):*
> ```cpp
> 
>     float d1_confidence = 0.65;
>     detections0.object_type_probs = { { p1_confidence, "person" }, {p2_confidence, "person" },
> ```

How can one detection have the same "person" classification listed twice? If Tiny YOLO is trained to classify 80 different things, then in practice won't a detection region have 80 unique classifications associated with it? If two people appear in an image, won't that result in two separate detections?

---

*[cpp/DarknetDetection/test/test_darknet_detection.cpp, line 286 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7eZQQCNOuI5D7eusaA:-L7eZQQCNOuI5D7eusaB:b-wx0vhz) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/test_darknet_detection.cpp#L286)):*
> ```cpp
>                                       { d1_confidence, "dog" } };
> 
>     {
> ```

Any reason for making these test cases separate code blocks other than readability?

---

*[cpp/DarknetDetection/test/test_darknet_detection.cpp, line 301 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7e_BlSgnHRtrt0XVBM:-L7e_BlSgnHRtrt0XVBN:b-3wxj5) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/test_darknet_detection.cpp#L301)):*
> ```cpp
> 
>     {
>         DarknetResult detections0OtherLocation;
> ```

Consider using underscores in this variable name for consistency and readability.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-:-L7ecMFty9vj8X_vRNVa:b-eg3asl)*

jrobble:
Reviewed 49 of 166 files at r1.
Review status: 109 of 166 files reviewed at latest revision, 21 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/52)*

brosenberg42:
I still need to make the .ini file / descriptor related changes.


> Which files in darknet_lib did you modify, if any?

I didn't modify any of the darknet source code.

> Is it possible to squelch printing out the network? It's not a big deal, but it is clutter.

This would require changes to the darknet source code.

---

Review status: 101 of 128 files reviewed at latest revision, 20 unresolved discussions.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 107 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aXShwkHOvHm367YMJ:-L8-2E2Gzn-NtkoD3wI1:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L107)):*

USE_PREPROCESSOR should work on images too. That should result in using the superset regions, just like it does for videos.
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 139 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aVBwSgMTjATW3-neU:-L8-2GG__-u_8x81wfcc:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L139)):*

I think this should be "CLASS" to match Caffe. I understand that this also returns the detection region, so it's different, but it is performing classification.
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 179 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7farUehTUuqlFDAu4U:-L8-2Jnox3yJvCcY56n2:b-fgakzp) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L179)):*

Why the +1 on `num_classes + 1`? Looking at this code:


```
      for (int name_idx = 0; name_idx < num_classes_; name_idx++) {
            float prob = probs_[box_idx][name_idx];
            if (prob >= confidence_threshold_) {
                detection.object_type_probs.emplace_back(prob, names_[name_idx]);
            }
        }
```

If there are 80 classes, won't `name_idx` be 0 ... 79?

I assume it has something to do with how `probs` is populated here:

```
    network_detect(network_.get(), image.dnet_image, confidence_threshold_, 0.5, 0.3, boxes_.get(),
                   probs_.Get());
```

A comment would be helpful.

</blockquote></details>

Added comment to source code.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 212 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fdZY3bDB_8GpIJcb2:-L8-2NslS18xtyRZjs6D:b-yv99rh) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L212)):*

Why must you set 0.01f as the minimum? What would happen if you set 0?
</blockquote></details>

Added comment to source code

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 224 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fd-tgHfBPVDwb7nmS:-L8-2RqwHCV4fk-cTy9e:b-yv99rh) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L224)):*

A comment explaining why we chose 0.5 and 0.3 would be helpful here.
</blockquote></details>

Added comment to source code

---

*[cpp/DarknetDetection/DarknetDetection.h, line 137 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fJKqEDZUgEA9MfbBg:-L8-2TnQLpALR2bpALdq:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.h#L137)):*

Let's call this "DarknetImageHolder". We don't shorten darknet anywhere else.
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/Trackers.cpp, line 83 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7eh0wrTZW2x5HfS_jr:-L8-2VxyjbUN5ouDd8Hq:b-v49cth) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L83)):*

The terms `other_confidences` and `other_types` is a little confusing. What is the "other" in this case? Those lists also include the top confidence and type.
</blockquote></details>

I changed the variable names

---

*[cpp/DarknetDetection/Trackers.cpp, line 94 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aKAA8zj_IIg4K-Mxx:-L8-2_XcFPkbH0JgqXas:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L94)):*

The Caffe component generates the following:

```
"detectionProperties": {
    "CLASSIFICATION": "golden retriever",
    "CLASSIFICATION CONFIDENCE LIST": "0.984752",
    "CLASSIFICATION LIST": "golden retriever"
},
```

For consistency, I think the Darknet component should use the same keys.
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/Trackers.cpp, line 102 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7ehZ4_xQ_NS31KYH5Y:-L8-2aln1BWJMOaYZ_zT:b-wws5wv) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L102)):*
> I think you're assuming that GetTracks() is called just before this class is destructed.

That is correct. It was just as easy to reset the state of the tracker to its initial state, so I did that. I added a comment in the header file mentioning that calling GetDetections resets the tracker.

---

*[cpp/DarknetDetection/Trackers.cpp, line 113 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fXXGwkCISiBse2UOQ:-L8-2vpTkxgNRQ8PFPfj:b-52t30s) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L113)):*

Shouldn't we try to combine image locations before adding a new image location to the track?

Because of the other you have these, it seems that if there exists a track that ends on the previous frame which has the same object type as our new detection, then that will always result in adding a new image location to the track.

Actually ... adding a new image location changes the key in the tracks map, so the next new detection in the loop will get combined. Never mind. It may be worth adding a comment here explaining this. 
</blockquote></details>

Added comment, changed order, and updated variable names.

---

*[cpp/DarknetDetection/Trackers.cpp, line 182 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7em6ovqd4KZk1TmOEE:-L8-3-BBeEmxsYFfovBj:bwzfgo0) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/Trackers.cpp#L182)):*

This `move` seems like it will leave the original tracks collection in a partially unspecified state.
</blockquote></details>

The elements of the collection were in an unspecified state. The collection itself was in a specified state. I added `tracks_.clear()` so that the moved from tracks won't stay around.

---

*[cpp/DarknetDetection/plugin-files/descriptor/descriptor.json, line 53 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aOYWahBYnzVzZzWg5:-L8-3S0jT_Ii3lU_9cxo:b-sybt5r) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/plugin-files/descriptor/descriptor.json#L53)):*
> or equal to the number of classifications that the model supports.

This is isn't true. It will just use all the classifications if the number is too high.

In `MPFImageLocation SingleDetectionPerTrackTracker::CreateImageLocation(int num_classes_per_region, DarknetResult &detection)` 
there is the line:  `int num_items_to_get = std::min(num_classes_per_region, static_cast<int>(object_probs.size()));`

---

*[cpp/DarknetDetection/test/CMakeLists.txt, line 1 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aP691jlV9ug6O6S9E:-L8-48KOkXpSrnv8Epc0:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/CMakeLists.txt#L1)):*

Missing header comment block.
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/test/test_darknet_detection.cpp, line 283 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7e_vNhKqGhXykWyCHA:-L8-49ewdWlNCbP-nHvf:bnnwu37) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/test_darknet_detection.cpp#L283)):*

How can one detection have the same "person" classification listed twice? If Tiny YOLO is trained to classify 80 different things, then in practice won't a detection region have 80 unique classifications associated with it? If two people appear in an image, won't that result in two separate detections?
</blockquote></details>

You are correct, that was a mistake.

---

*[cpp/DarknetDetection/test/test_darknet_detection.cpp, line 286 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7eZQQCNOuI5D7eusaA:-L8-4EEmgqNLw3yft9ms:b-6vdrt) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/test_darknet_detection.cpp#L286)):*

Any reason for making these test cases separate code blocks other than readability?
</blockquote></details>

That was so that a fresh tracker gets created in each block.

---

*[cpp/DarknetDetection/test/test_darknet_detection.cpp, line 301 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7e_BlSgnHRtrt0XVBM:-L8-4KNWVYFy3IlhTCor:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/test_darknet_detection.cpp#L301)):*

Consider using underscores in this variable name for consistency and readability.
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/test/data/NOTICE, line 5 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aQ4Maar0JvXdqEPIA:-L8-4MA4uvBFHIh6awFf:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/data/NOTICE#L5)):*

For completeness, also list where dog.jpg came from.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-:-L8-1eiWJxEajO549P7W:b5nx9v7)*

brosenberg42:
Review status: 101 of 128 files reviewed at latest revision, 20 unresolved discussions.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 115 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aS_u9VyyfWYfgR-eZ:-L82f9n46tZQqrfubACT:b-nqvmyh) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L115)):*

Instead of hardcoding the default value, the convention is to use an .ini file.
</blockquote></details>

That was before we started passing all the job properties, including the unspecified default properties, to the components. One of the reasons we decided to do that was because the property values in the .ini files may not match the property values specified in the descriptor file. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/52)*

jrobble:
>> Is it possible to squelch printing out the network? It's not a big deal, but it is clutter.

> This would require changes to the darknet source code.

I think we should do it, or are you opposed to modifying their source code? I see no reason why not. We have the freedom.

---

Reviewed 5 of 54 files at r2.
Review status: 106 of 128 files reviewed at latest revision, 8 unresolved discussions.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 115 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aS_u9VyyfWYfgR-eZ:-L82pou8rIrLKlxKVrWl:bchk32c) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L115)):*

That was before we started passing all the job properties, including the unspecified default properties, to the components. One of the reasons we decided to do that was because the property values in the .ini files may not match the property values specified in the descriptor file. 
</blockquote></details>

Ok. Please disregard.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 212 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fdZY3bDB_8GpIJcb2:-L82qI3neTDKcmKWVnI3:bmasui2) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L212)):*

Added comment to source code
</blockquote></details>

This prevents someone from configuring the job so that it reports every possible classification. I don't think we want to preclude someone from doing that.

---

*[cpp/DarknetDetection/Trackers.cpp, line 113 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L82tgFQ-rYxbjPQd7Gt:-L82tgFQ-rYxbjPQd7Gu:b-9qtv1y) ([raw file](https://github.com/openmpf/openmpf-components/blob/355b51d9ca38e35ea4f2df7b868dcc1f9481e73a/cpp/DarknetDetection/Trackers.cpp#L113)):*
Please add "one":

> // Check if there is more than one box ...

---

*[cpp/DarknetDetection/plugin-files/descriptor/descriptor.json, line 53 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aOYWahBYnzVzZzWg5:-L82rPoj0FOu10mhJ9zq:bkzybe9) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/plugin-files/descriptor/descriptor.json#L53)):*
Ok. Then how about we add:

> At most, the maximum number of classifications that the model supports will be reported.

---

*[cpp/DarknetDetection/test/test_darknet_detection.cpp, line 286 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7eZQQCNOuI5D7eusaA:-L82s3yQ302tDx1g1-kj:b-phdquz) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/test/test_darknet_detection.cpp#L286)):*

That was so that a fresh tracker gets created in each block.
</blockquote></details>

Oh, I see. Works for me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-:-L82v88CdWJsiPILjz9v:bq12cjp)*

brosenberg42:
> I think we should do it, or are you opposed to modifying their source code? I see no reason why not. We have the freedom.

I prefer not to. It will make it easier if we decided to update to a newer version of Darknet.

---

Review status: 106 of 128 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/DarknetDetection/DarknetDetection.cpp, line 212 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7fdZY3bDB_8GpIJcb2:-L83QVp_OkN-L6GmnJu9:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/DarknetDetection.cpp#L212)):*

This prevents someone from configuring the job so that it reports every possible classification. I don't think we want to preclude someone from doing that.
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/Trackers.cpp, line 113 at r2](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L82tgFQ-rYxbjPQd7Gt:-L83QXq5Lul3r5K0Uobb:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/355b51d9ca38e35ea4f2df7b868dcc1f9481e73a/cpp/DarknetDetection/Trackers.cpp#L113)):*

Please add "one":

> // Check if there is more than one box ...
</blockquote></details>

Done.

---

*[cpp/DarknetDetection/plugin-files/descriptor/descriptor.json, line 53 at r1](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-L7aOYWahBYnzVzZzWg5:-L83QaNDImupoA5dtsMt:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/plugin-files/descriptor/descriptor.json#L53)):*

Ok. Then how about we add:

> At most, the maximum number of classifications that the model supports will be reported.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf-components/52#-:-L83Lb52jCmtqBGc_5sf:borvv6u)*

jrobble:
Since that output only shows up when the gtests are run, it's not a big deal. It doesn't clutter the log output. Let's not worry about it.

---

Reviewed 15 of 166 files at r1, 7 of 54 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/DarknetDetection/plugin-files/config/data/9k.names, line 2410 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf-components/52#-L83LaUoRIU1qD6EMUEY:-L83LaUqpXY_O_Rk9i4H:b-e5wc19) ([raw file](https://github.com/openmpf/openmpf-components/blob/355b51d9ca38e35ea4f2df7b868dcc1f9481e73a/cpp/DarknetDetection/plugin-files/config/data/9k.names#L2410)):*
> ```
> spar
> mast
> mast
> ```

It's interesting that mast is in here twice.

---

*[cpp/DarknetDetection/plugin-files/descriptor/descriptor.json, line 26 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf-components/52#-L7aMwpWJ1BevI3fjCki:-L83beCAB2UaVI5HXRqN:b-2a9u5m) ([raw file](https://github.com/openmpf/openmpf-components/blob/17382952ec6715c15465bdc129eeec38d505a0b5/cpp/DarknetDetection/plugin-files/descriptor/descriptor.json#L26)):*

I see the following unique properties:

- NETWORK_CONFIG_FILE
- WEIGHTS_FILE
- NAMES_FILE
- USE_PREPROCESSOR
- NUMBER_OF_CLASSIFICATIONS_PER_REGION

The convention is to list the unique properties after the list of common properties. That makes it easier to compare the descriptor files for consistency.

Instead of requiring the user to specify three properties (NETWORK_CONFIG_FILE, WEIGHTS_FILE, NAMES_FILE), let's consider creating a .json file that serves the same purpose as openmpf-components/cpp/CaffeDetection/plugin-files/config/models.cfg. That file groups the models and their associated files together so then all the user needs to do is specify the value of MODEL_NAME.

</blockquote></details>

I created https://github.com/openmpf/openmpf/issues/512 for later.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf-components/52#-:-L83dot4ATXD-ZyPKL7S:brboad8)*

PR #64, opened 2018-06-14
Add FRAME_RATE_CAP to descriptors

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/64)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Included in this repository are the following C++ Components:
 - DLIB Face Detection
 - OpenCV LBP Face Detection
 - OpenALPR Text Detection
+- Darknet Object Classification
 
 And the following Java Components:
 - Sphinx Speech Detection
--- a/cpp/CaffeDetection/plugin-files/descriptor/descriptor.json
+++ b/cpp/CaffeDetection/plugin-files/descriptor/descriptor.json
@@ -44,9 +44,15 @@
         },
         {
           "name": "FRAME_INTERVAL",
-          "description": "Controls whether the component performs detection on every frame in the video segment, or skips some frames at a regular interval. Must be greater than or equal to 0. If the frame_interval is set to 0 or 1, a frame_interval of 1 will be used, so that detections are performed on every frame. For a frame interval N > 1, every N-1 frames will be skipped. Default value is defined by the MPF properties file.",
+          "description": "Controls whether the component performs detection on every frame in the video segment, or skips some frames at a regular interval. Must be set to a value >= 0. If set to 0 or 1, a frame interval of 1 will be used, meaning that detection is performed on every frame. If set to N > 1, every N-1 frames will be skipped. Default value is defined by the OpenMPF properties file.",
           "type": "INT",
           "propertiesKey": "detection.sampling.interval"
+        },
+        {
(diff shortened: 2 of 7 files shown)


jrobble:
Update Sphinx too.

---

Reviewed 6 of 6 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/64#-:-LF5JFvZAJgpQxsgMF_P:b-p1q3uo)*

brosenberg42:
For Sphinx, I understand that you want to change MPF properties file to OpenMPF properties file. Currently the Sphinx descriptor has a FRAME_INTERVAL property. Should that even be there?

---

Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/64#-:-LFHj_zD7yBAEEk9MZDt:b3p3uqi)*

jrobble:
Here are the current Sphinx properties:

1. FRAME_INTERVAL
2. CONFIDENCE_THRESHOLD
3. MIN_GAP_BETWEEN_SEGMENTS
4. TARGET_SEGMENT_LENGTH
5. MIN_SEGMENT_LENGTH
6. MERGE_TRACKS
7. MIN_GAP_BETWEEN_TRACKS
8. MIN_TRACK_LENGTH
9. MIN_OVERLAP

Regarding how these are used by the WFM:

1 is used to create a TrackMergingPlan, but that field getter is never used. Also, it is used to create a SegmentingPlan, but that field getter is never used. For all intents and purposes, it seems that FRAME_INTERVAL is not used by the WFM for anything; however, at one time we had considered using it to create segments that start/stop on frame intervals.

2 applies to all media types. It's used to filter out detections with low confidence.

3-5 are used by the DetectionSplitter to create a SegmentingPlan, which is only done for VIDEO. However, Sphinx can process video, so technically, in that case, multiple Sphinx jobs are created for a single piece of video media.

6-9 are used by the TrackMergingProcessor, which the DetectionResponseRouteBuilder applies to every detection response. It really only makes sense for videos, so it should probably be disabled for Sphinx tracks.

To answer your question, Sphinx can process videos, and the FRAME_INTERVAL is part of the TrackMergingPlan and SegmentingPlan (even though it isn't currently used for anything, we might eventually use it for segmenting). That said, let's include FRAME_INTERVAL and FRAME_RATE_CAP for Sphinx. If Sphinx was only processing audio files, then it would be a different story.

While you're at it, please set MERGE_TRACKS to FALSE in the Sphinx descriptor file (it currently uses the system property). Also, remove "Default value is defined by the MPF properties file" from the description.

---







*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/64#-:-LFHl_xZ5jN0PjM572u2:b-x6m6r8)*

jrobble:
Reviewed 1 of 1 files at r2.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/64)*

PR #67, opened 2018-06-28
Add tracking to Darknet.

brosenberg42:


jrobble:
Review status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @brosenberg42 and @jrobble)

---

*[cpp/DarknetDetection/plugin-files/descriptor/descriptor.json, line 202 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/67#-LGQwkLt79Xi91R8z1Al:-LGQwkLu1VYxpikjb0nz:b-z99ze2) ([raw file](https://github.com/openmpf/openmpf-components/blob/6dcd725709455d8c15a0c9b4a8372e41eb1d2ad8/cpp/DarknetDetection/plugin-files/descriptor/descriptor.json#L202)):*
> ```JSON
>         },
>         {
>           "name": "RECTANGLE_INTERSECTION_MIN",
> ```

There are three things to consider here:

1. OCV face detection hardcodes this value to [0.15 in one place](https://github.com/openmpf/openmpf-components/blob/431a01171479ea2f51710aba54923e5592d361bc/cpp/OcvFaceDetection/OcvFaceDetection.cpp#L278), but it uses [0.7 in another](https://github.com/openmpf/openmpf-components/blob/431a01171479ea2f51710aba54923e5592d361bc/cpp/OcvFaceDetection/OcvFaceDetection.cpp#L685) to match face templates. We certainly shouldn't hardcode it. The thing to note is that although 0.15 seems really low, they are during more than pure region overlap, so it's hard to get a handle on just what that value means.

2. Dlib face detection has a MAX_INTERSECTION_OVERLAP_AREA_PCT property that's never used. See [this code](https://github.com/openmpf/openmpf-components/blob/master/cpp/DlibFaceDetection/DlibFaceDetection.cpp#L308). Any overlap at all results in two detections being part of the same track. However, dlib does choose the track with the greatest overlap % in [this code](https://github.com/openmpf/openmpf-components/blob/master/cpp/DlibFaceDetection/DlibFaceDetection.cpp#L279).

3. Every component that processes video declares the MIN_OVERLAP property in its descriptor.json file. Is there a reason to use a separate RECTANGLE_INTERSECTION_MIN property here, or can the same MIN_OVERLAP property be used? That property has a default value defined by detection.track.overlap.threshold, which is 0.60. To me, it seems to make the most sense that the component uses the same property as the WFM. However, thinking about this more, the WFM will combine tracks with no regard for the CLASSIFICATION. To address that, I see two options:

   1. Set MERGE_TRACKS to false here.

   2. Update TrackMergingProcessor.combine() to check the CLASSIFICATION if the track type is CLASS before attempting to merge tracks.

Depending on what we decide for # 3, [this code](https://github.com/openmpf/openmpf/blob/9582fb2e1b633c3ec9eac189553400f371f5e3c5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/trackmerging/TrackMergingProcessor.java#L289) needs to be updated. Minimally, "SCENE" needs to be added to that list. It would be better if the values were read from a .properties file.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/67)*

PR #71, opened 2018-07-27
Frame queue

clnowacki:


brosenberg42:
___
*[cpp/DarknetDetection/darknet_wrapper/DarknetImpl.h, line 37 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/71#-LISR7Bb-9TkX5xJ9o8S:-LIf-Uud4mR4MBrERYYo:b-2bi3ve) ([raw file](https://github.com/openmpf/openmpf-components/blob/b6b0511007d6f1f13f91206b1e271e9843dba1a5/cpp/DarknetDetection/darknet_wrapper/DarknetImpl.h#L37)):*

DarknetInterface.h now includes MPFDetectionComponent.h.
</blockquote></details>

Relying on transitive includes is considered bad practice unless the you are in a .cpp file and the include is in a .h file with the same name (e.g. foo.cpp and foo.h).

brosenberg42:
Jeff, the results you are seeing does not surprise me. If the consumer is slower than the producer, I would not expect that making the queue bigger would improve performance. Once you are in a situation where there is always at least one frame available as soon as the consumer tries to retrieve it, making the queue bigger shouldn't make a difference. 

Comparing these two situations, A) there is always one frame available when the producer calls pop, versus, B) there are always 10 frames available when the producer calls pop. Since the consumer only processes one frame at a time situation A and B are effectively the same thing.

PR #95, opened 2019-04-24
East component

cdglasz:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/95)


jrobble:
___
*[python/EASTTextDetection/east_component/east_component.py, line 65 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/95#-Ldy3-T34WUgKKhWPa54:-LeSCRgUDRgmeIVWyzSv:b-wc1dyo) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5424582b4144953608645ac90bbcd44123cb23c/python/EASTTextDetection/east_component/east_component.py#L65)):*

There is a bug that seems related to caching between job runs. When I run two video jobs back-to-back in OpenMPF on the same media, and change the BATCH_SIZE from 1 to 32 in the second job, the output of the second job is missing detections in some frames. Those detections are present in the output from the first job.

When I stop the service through the Nodes UI, start it again, and then run the job with BATCH_SIZE=32 the output looks correct. Stopping and starting the service seems to be clearing something that got unintentionally persisted between job runs.
</blockquote></details>

Also, the second job is processed very fast.

jrobble:
___
*[python/EASTTextDetection/east_component/east_component.py, line 65 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/95#-Ldy3-T34WUgKKhWPa54:-LeSWgEg5Ad5IzFnW-VI:b-f0des6) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5424582b4144953608645ac90bbcd44123cb23c/python/EASTTextDetection/east_component/east_component.py#L65)):*

Also, the second job is processed very fast.
</blockquote></details>

Further testing shows this only happens when the last batch of frames is smaller then the others. For example, using a video with 68 frames, a batch size of 1, 2, or 4 is fine. Sizes of 3, 5, 16, and 32 are not.

jrobble:
___
*[python/EASTTextDetection/east_component/east_component.py, line 65 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/95#-Ldy3-T34WUgKKhWPa54:-LeSbnOo0nxrM5yovLTA:b2sdg6u) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5424582b4144953608645ac90bbcd44123cb23c/python/EASTTextDetection/east_component/east_component.py#L65)):*

Further testing shows this only happens when the last batch of frames is smaller then the others. For example, using a video with 68 frames, a batch size of 1, 2, or 4 is fine. Sizes of 3, 5, 16, and 32 are not.
</blockquote></details>

More testing shows that the frames skipped in the second job correspond to a pattern. For example, if the batch size is 5 and the last batch has 3 frames, then when the second job is processed the first 3 frames have detections and the next 2 don't. The next 3 have detections and the next 2 don't, and so on. The 2 seems to be coming from how the last batch for the for job was missing the last 2 out of the 5 frames required for a full batch.

FYI, this is the video I'm using for testing: https://github.com/openmpf/openmpf/blob/master/trunk/mpf-system-tests/src/test/resources/samples/text/ff-region-motion-lp.avi

jrobble:
___
*[python/EASTTextDetection/east_component/east_component.py, line 65 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/95#-Ldy3-T34WUgKKhWPa54:-LeSfWguEHO4HsUzQetk:bnsjexa) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5424582b4144953608645ac90bbcd44123cb23c/python/EASTTextDetection/east_component/east_component.py#L65)):*
May be related to this: https://github.com/opencv/opencv/issues/13462

I've read:

>   Varying input blobs' sizes such height or width or batch size you make OpenCV reallocate all the internal memory. That leads efficiency gaps. Try to initialize and deploy models using a fixed batch size and image's dimensions.

And I've read that keeping the batch size consistent is the responsibility of the user: https://github.com/opencv/opencv/issues/10210#issuecomment-348454604

We certainly want to process all of the frames of a video. Adding dummy frames to ensure a full batch seems wasteful. It seems that it may be safest to clear the models after processing a video to ensure we don't run into this issue. 

I don't want to be in the business of counting frames and batch remainders to see if we can use the same models between video jobs. I'm not confident that we can rely on the state of the Net if something goes wrong, like an exception occurs or we can't read a frame.

jrobble:
___
*[python/EastTextDetection/east_component/east_processor.py, line 337 at r9](https://reviewable.io/reviews/openmpf/openmpf-components/95#-LeSPYjG36MtCL66F-tV:-LeSPYjHE-DrekDoCz6x:b-rn4ybs) ([raw file](https://github.com/openmpf/openmpf-components/blob/5d900ea1ed72a24bc28423cd8c9d71fb6835cc45/python/EastTextDetection/east_component/east_processor.py#L337)):*
> ```Python
>             rotate_on=rotate_on
>         )
> ```

Say "Convert the images ..."

jrobble:
___
*[python/EastTextDetection/setup.py, line 35 at r10](https://reviewable.io/reviews/openmpf/openmpf-components/95#-LfV0UmXEL5iLt6yZvda:-LfV0UmXEL5iLt6yZvdb:b-wqfdrg) ([raw file](https://github.com/openmpf/openmpf-components/blob/1ea8a76e3ea9bd5453767dcafd2022e76af2890f/python/EastTextDetection/setup.py#L35)):*
> ```Python
>     packages=setuptools.find_packages(exclude=('*test*',)),
>     package_data={'': ['east_resnet50.pb']},
>     install_requires=(
> ```

Please add `opencv-python>=3.3` to this list for good measure.

PR #98, opened 2019-05-30
Tesseract component now calls ReverseTransform().

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/98)


jrobble:
I decided to split out the descriptor updates to [this branch](https://github.com/openmpf/openmpf-components/tree/tesseract-east-descriptor). We need to wait until @cdglasz lands the EAST component before merging those changes.

PR #103, opened 2019-06-13
Fixed: Tesseract now updates upper left coordinates for image rotatio…

hhuangMITRE:
…ns detected by OSD mode.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/103)


brosenberg42:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1523 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/103#-LhMUTztDHp8KFMkhYkJ:-LhMUTzu-3vqxrwkOpq2:b-u0ix05) ([raw file](https://github.com/openmpf/openmpf-components/blob/a6531e97d38794495d10a02fb7a0ce872c195ecc/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1523)):*
> ```cpp
>                 case 1:
>                     // 90 degree counterclockwise rotation.
>                     yLeftUpper = input_size.height - 1;
> ```

For the 90 degree and 270 degree rotations, we also need to swap the width and the height of the detection.

jrobble:
> Do you also want me to swap width and height for the MPFImageLocation parameter when we apply 90, 270 degree rotations? Or did you comment reflect the fix that was just applied?

I will update your PR to do this. It's necessary; other wise we get:

odd

PR #108, opened 2019-07-03
Adding Two Pass OCR Processing for Rotated Text. Also updating README…

hhuangMITRE:
… documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/108)


jrobble:
___
*[cpp/TesseractOCRTextDetection/README.md, line 85 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/108#-LiuXxGI9tCS_h3lgpMf:-LiuXxGI9tCS_h3lgpMg:bc4v0km) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5e80baa184aa9d7765c56f6812a9c8bacc62fa0/cpp/TesseractOCRTextDetection/README.md#L85)):*
> ```Markdown
> * After performing two-pass OCR, the output with the highest OCR confidence is kept as the final track.
> * For multiple, separate language tracks, the previous steps are repeated for each specified language track, such that one high confidence output for each track is returned by the component.
> * To distinguish which track (first pass vs second pass) is returned, the component will also specify the optimal rotation in TWO_PASS_OCR_ROTATION, alongside the original ROTATION property detected by OSD.
> ```

Brian's arbitrary rotation code will only look at the ROTATION detection property when performing a reverse transform. All components should standardize on outputting ROTATION.

For example, if adding 180 to OSD's rotation value produces the best results, your component should set the ROTATION detection property to OSD's rotation + 180. The way to test this is to make sure the colored dot in the OpenMPF markup is in the correct location (the upper-left of the bounding box if it was correctly rotated, which would be the lower right in the original image).

I understand your intent with TWO_PASS_OCR_ROTATION detection property. Without it, if we set ROTATION to OSD's rotation + 180, and leave TEXT_ORIENTATION_CONFIDENCE set to OSD's rotation confidence, then it would appear that OSD correctly identified the rotation.

Instead of TWO_PASS_OCR_ROTATION, let's use OSD_TEXT_ORIENTATION_CORRECTION=180. Leave that set to 0 if the first pass was better. To be consistent, I think we should rename TEXT_ORIENTATION_CONFIDENCE to OSD_TEXT_ORIENTATION_CONFIDENCE (that will also better match what you have now with MIN_OSD_TEXT_ORIENTATION_CONFIDENCE).

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1024 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/108#-LiuXDZEA08YpJHAtZ-1:-LiuXDZEA08YpJHAtZ-2:b-ism7wb) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5e80baa184aa9d7765c56f6812a9c8bacc62fa0/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1024)):*
> ```cpp
>     if (detection_properties["PRIMARY_SCRIPT"] == "NULL") {
>         //If OSD failed to detect any scripts, automatically set confidence scores to -1 (not enough text).
>         detection_properties["PRIMARY_SCRIPT_CONFIDENCE"] = -1;
> ```

Shouldn't we also set PRIMARY_SCRIPT_SCORE=-1?

Since we plan to rename TEXT_ORIENTATION_CONFIDENCE to OSD_TEXT_ORIENTATION_CONFIDENCE, I think we should rename the other OSD detection properties as well:

- OSD_PRIMARY_SCRIPT
- OSD_PRIMARY_SCRIPT_CONFIDENCE
- OSD_PRIMARY_SCRIPT_SCORE
- OSD_SECONDARY_SCRIPTS
- OSD_SECONDARY_SCRIPT_SCORES

jrobble:
___
*[cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json, line 299 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/108#-LiugBNoDKotzLXG_0XZ:-LiugBNoDKotzLXG_0X_:b-em9i60) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5e80baa184aa9d7765c56f6812a9c8bacc62fa0/cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json#L299)):*
> ```JSON
>         },
>         {
>           "name": "ROTATE_AND_DETECT_MIN_CONFIDENCE",
> ```

Let's call this ROTATE_AND_DETECT_MIN_OCR_CONFIDENCE.

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1619 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/108#-LiugWWO-kWzGmsmiEWO:-LiugWWO-kWzGmsmiEWP:b7llbwl) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5e80baa184aa9d7765c56f6812a9c8bacc62fa0/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1619)):*
> ```cpp
>             final_out.two_pass_rotation = first_pass_rotation;
> 
>             if (ocr_fset.rotate_and_detect && (min_ocr_conf <= 0 || ocr_out.confidence < min_ocr_conf)) {
> ```

I think you want: `ocr_out.confidence <= min_ocr_conf`

jrobble:
___
*[cpp/TesseractOCRTextDetection/README.md, line 85 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/108#-LjMZNk2Dcw6iy68q4xX:-LjMxQtK3TMyEDVRtnoZ:b5qxjoj) ([raw file](https://github.com/openmpf/openmpf-components/blob/ff288ac1a9f0bdfe7baf80c9564ad7cc5e417552/cpp/TesseractOCRTextDetection/README.md#L85)):*

Let's rephrase:

> To distinguish which track (first pass vs second pass) is returned, the component will also specify the optimal rotation in OSD_TEXT_ORIENTATION_CORRECTION, alongside the ROTATION property detected by OSD. The ROTATION property will be updated by the value specified in OSD_TEXT_ORIENTATION_CORRECTION (either 0 or 180 degree rotation added to existing rotation value).

To:

> The component will report TEXT_ORIENTATION_PASS alongside the ROTATION property. The ROTATION property is the rotation of the text region, taking into consideration both the OSD rotation (if OSD was performed), as well as the 2-pass rotation attempt. TEXT_ORIENTATION_PASS is set to 0 or 180 depending on which pass produced the best results.

Sorry to flip-flop. After reading your description, I realized that the two-pass may be performed even when OSD automation is disabled, thus we should rename the detection property from OSD_TEXT_ORIENTATION_CORRECTION to TEXT_ORIENTATION_PASS.
</blockquote></details>

Actually, use ROTATE_AND_DETECT_PASS instead of TEXT_ORIENTATION_PASS. That we way can standardize on ROTATE_AND_DETECT_PASS across components.

PR #109, opened 2019-07-03
Customized OSD for feed-forward regions to be more aggressive in iden…

hhuangMITRE:
…tifying language and orientation. Increased rescale parameter for input regions.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/109)

### DIFF
--- a/cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json
+++ b/cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json
@@ -13,7 +13,7 @@
   ],
   "algorithm": {
     "name": "TESSERACTOCR",
-    "description": "This component applies Tesseract optical character recognition to an image.",
+    "description": "Performs Tesseract optical character recognition.",
     "actionType": "DETECTION",
     "requiresCollection": {
       "states": []
@@ -295,7 +295,7 @@
   "actions": [
     {
       "name": "TESSERACT OCR TEXT DETECTION ACTION",
-      "description": "Runs the Tesseract OCR model on an image.",
+      "description": "Performs Tesseract OCR.",
       "algorithm": "TESSERACTOCR",
       "properties": [
         {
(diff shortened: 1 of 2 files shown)


jrobble:
___
*[cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json, line 336 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/109#-LiuRV7vAxi5omTDlx_i:-LiuQt6A94p8Ya1vGPJT:b-ehu7ds) ([raw file](https://github.com/openmpf/openmpf-components/blob/ba01b7bdfd6cbc85319b6e3ca34182b707afeeea/cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json#L336)):*
> ```JSON
>         },
>         {
>           "name": "MIN_OSD_ROTATION_CONFIDENCE",
> ```

Note that when defining an action, you only need to specify the name and value for the properties. The description and default value are only needed when defining the properties for the algorithm. You can think of an action as an instantiation of an algorithm. You give the action specific property values that override the algorithm defaults.

I will be updating this PR shortly to correct this and move the feed-forward pipelines to EAST to prevent a component registration dependency issue.

jrobble:
___
*[cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json, line 356 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/109#-LjH_W8_CXjHqQIkDfwl:-LjMVpFC8urX_M_ef95d:b-sdwrf1) ([raw file](https://github.com/openmpf/openmpf-components/blob/e41b714c45682bd9ca7f56e9acbf41813fdffb5f/cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json#L356)):*

Got it, I've wrapped up all of the PRs for the wild text branch and the two-pass OCR branch.

Once those two PRs look ready to land, I'll start to land them one by one into develop, or merge all three into this branch if that works better for tracking and review.

In the meantime I will begin merging each of the PRs together in a copy of this branch. Let me know if there are any additional PR edits needed for the other two branches, thanks!


</blockquote></details>

I think it makes the most sense to land them to develop, one by one. It will be easier to catch merge conflicts that way.

PR #111, opened 2019-07-10
Tesseract remove gibberish filters.

hhuangMITRE:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/111)

### DIFF
--- a/cpp/TesseractOCRTextDetection/README.md
+++ b/cpp/TesseractOCRTextDetection/README.md
@@ -14,9 +14,7 @@ Please refer to https://imagemagick.org/script/formats.php for support of other 
 Users may set the language of each track using the TESSERACT_LANGUAGE parameter
 as well as adjust image preprocessing settings for text extraction.
 
-For processing English text, users can enable filters (THRS_FILTER, HIST_FILTER)
-to eliminate gibberish detections from a given scene. All text extracted from
-an image can also be tagged using regex and keyword tags in a given json file.
+All text extracted from an image can also be tagged using regex and keyword tags in a given json file.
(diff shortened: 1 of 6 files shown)


jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1512 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/111#-LjSnoan24DwvF8tWf0m:-LjSoo4OBraUwh5Fe5lF:b-qapzb6) ([raw file](https://github.com/openmpf/openmpf-components/blob/7ead39cd59d4331c8c0d42b1bee0d622692b5814/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1512)):*

Reading the description: 

```
          "name": "ROTATE_AND_DETECT_MIN_OCR_CONFIDENCE",
          "description": "Specifies minimum text confidence required to skip second OCR pass with 180 degree rotation. Negative values will result in ROTATE_AND_DETECT performing two pass OCR regardless of text confidence from the initial pass.",
          "type": "DOUBLE",
          "defaultValue": "95.0"
```

I think I told you the wrong thing before. I think we do want `ocr_out.confidence < min_ocr_conf` here. That means we should change the following in the README:

```
    * If the first pass is greater than the minimum specified threshold, the second pass of the OCR is skipped.
```

to:

```
    * If the first pass is less than the minimum specified threshold, the second pass of the OCR is skipped.
```

</blockquote></details>

Err, I mean we should change it to:

```
    * If the first pass is greater than or equal to than the minimum specified threshold, the second pass of the OCR is skipped.
```

PR #112, opened 2019-07-15
Dockerize OCV face

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/112)


brosenberg42:
___
*[cpp/OcvFaceDetection/Dockerfile, line 37 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/112#-LkZg3eqCg8paBuUcFCH:-LkZoCr97AICUTJnbWQB:bpahfmn) ([raw file](https://github.com/openmpf/openmpf-components/blob/ac0fa3cd4d0548c82d2bab4fab9d868fc02c567e/cpp/OcvFaceDetection/Dockerfile#L37)):*
> When not just use "${RUN_TESTS}" == true? I'm not familiar with the ,, part.

That is to make the comparison case insensitive.

Both comments about the caching are correct.

PR #116, opened 2019-08-19
Added case-sensitive regex tag processing. Updated trigger words repo…

hhuangMITRE:
Added case-sensitive regex tag processing. Updated trigger words reporting behavior (now returns original text rather than normalized lowercase text) and updated test cases.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/116)


jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1055 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/116#-Lo7Jno54_HoJ-mwl7Hg:-LoMImES4qOVC_xiIGHY:b-ucpf4l) ([raw file](https://github.com/openmpf/openmpf-components/blob/2f155a71cf17741e3290e162bf11a3e9d1019ad4/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1055)):*

After reviewing your Tika text detection PR, I have an idea. Instead of converting the full text and regex pattern to lowercase yourself, instead use `boost::regex::icase` when ignoring case (when `[[:case_sensitive:]]` is not specified). For example:

```
static const boost::regex bad_words("...your regex...", boost::regex::icase);
```

This is similar to using the following in the Tika text detection component:

```
Pattern.compile("Registrar:\\s(.*)", Pattern.CASE_INSENSITIVE);
```

Following this approach, I don't think you will need to convert anything to lowercase and all of the Boost regex operators will work as expected.
</blockquote></details>

Thinking about this a bit more, and how the use of `[[:case_sensitive:]]` doesn't really work well for the Tika text detection component, I'm beginning to wonder if we should instead just change the form of the JSON:

```
  "TAGS_BY_REGEX": {
    "vehicle": [
      {"pattern": "auto", "caseSensitive": false},
      {"pattern": "(\\b)bike(\\b)", "caseSensitive": false},
      {"pattern": "(\\b)bus(\\b)", "caseSensitive": false}
    ],
```

Then we don't have to worry about this issue.

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1055 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/116#-Lo7Jno54_HoJ-mwl7Hg:-LoMQJnqAMI-Gn1bctxM:b-rxq509) ([raw file](https://github.com/openmpf/openmpf-components/blob/2f155a71cf17741e3290e162bf11a3e9d1019ad4/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1055)):*

Great idea, I was also thinking of changing to setting the case insensitivity flag  rather than the original lowercase conversion. 

The JSON changes make sense. Perhaps we can specify caseSensitive as false by default so users only need to set the caseSensitive flag to true when they want certain patterns to be case sensitive. Should we also maintain backwards compatibility with the original json tag format? Kind of like:

```
 "TAGS_BY_REGEX": {
    "vehicle": [
      "auto",
      "(\\b)bike(\\b)",
      {"pattern": "(\\b)bus(\\b)", "caseSensitive": true}
    ],

```
In this case, both `auto` and `bike` tags behave like the original tags (case insensitive by default), but the tag for `bus` is now case sensitive.


 
</blockquote></details>

I like the idea of you presented above, but I don't think that's valid JSON. We're going from an array of strings to an array of objects, which means I think each object has to have the key-value pair mapping for each of the object properties.

I know in Java you can allow some of the object properties to take on a default value, so in that case the best we can do is something like:

```
  "TAGS_BY_REGEX": {
    "vehicle": [
      {"pattern": "auto"},
      {"pattern": "(\\b)bike(\\b)"},
      {"pattern": "(\\b)bus(\\b)", "caseSensitive": true}
    ],
```

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1055 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/116#-Lo7Jno54_HoJ-mwl7Hg:-Lofy73W7HrC0tuOcBF_:b4gjqbs) ([raw file](https://github.com/openmpf/openmpf-components/blob/2f155a71cf17741e3290e162bf11a3e9d1019ad4/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1055)):*

Got it, thanks. I'll update and test the JSON changes in both branches and include them in the next update for both.
</blockquote></details>

Ok. Sounds good.

PR #119, opened 2019-08-27
Updating Tika Component to follow Tesseract trigger words and regex tagging behavior.

hhuangMITRE:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/119)


jrobble:
___
*[java/TikaTextDetection/plugin-files/config/text-tags.json, line 90 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/119#-LoLzJneD52NGgKKZeBb:-LoMJSejEqQCifd8Pywa:b-qv576g) ([raw file](https://github.com/openmpf/openmpf-components/blob/f800af7bf776a24529c8dfb2ddd31ebaf58c721b/java/TikaTextDetection/plugin-files/config/text-tags.json#L90)):*

It's inconsistent that we need to escape the square brackets for the delimiter tag above, but not escape the square brackets in `[[:case_sensitive:]]`.

I know the reason is that you find and remove `[[:case_sensitive:]]` before converting it to a Java regex, but from a user's point of view, this inconsistency is a little odd.

That said, it seems that should use `\\[\\[:case_sensitive:\\]\\]`; however, the only reason we're using `[[:case_sensitive:]]` to begin with is to be consistent with Tesseract, which we are now diverting from by adding slashes.

So that makes we wonder if we should just do something that more consistent with way Java regex specifies character classes, like using `{CaseSensitive}` at the beginning of the regex pattern.
</blockquote></details>

Thinking about this more, see this comment: https://github.com/openmpf/openmpf-components/pull/116#issuecomment-529610982

jrobble:
___
*[java/TikaTextDetection/plugin-files/descriptor/descriptor.json, line 2 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/119#-LoMNB8X9OiMMZ2sq5jc:-LoMNB8X9OiMMZ2sq5jd:b-4l4lyj) ([raw file](https://github.com/openmpf/openmpf-components/blob/f800af7bf776a24529c8dfb2ddd31ebaf58c721b/java/TikaTextDetection/plugin-files/descriptor/descriptor.json#L2)):*
> ```JSON
> {
>   "componentName" : "TikaTextDetection",
> ```

You've added the FULL_REGEX_SEARCH to the tesseract component, but it's lacking here. We should add it for consistency.

PR #126, opened 2019-10-21
Tesseract parallel processing

hhuangMITRE:
Adding parallel OCR processing for Tesseract component.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/126)


jrobble:
___
*[cpp/TesseractOCRTextDetection/README.md, line 189 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/126#-LuF1_qc8uR6sm7NmVfR:-LuIiHb85io9UqZeQsFc:b-4pyifs) ([raw file](https://github.com/openmpf/openmpf-components/blob/e40cd33a3d422ee8bb3f17e7f78af66f870ef424/cpp/TesseractOCRTextDetection/README.md#L189)):*

Oh, I meant to say that if any number of OSD-detected models are missing, then the default language automatically replaces those missing scripts. (ex. OSD -> run Latin and Cyrillic, but Latin is missing -> run English (default) and Cyrillic instead.)

I updated that statement alongside the job failure description.
</blockquote></details>

Ok. That behavior is different from the speech detection component, which will only use the user-provided models as a last resort when none of the automatically detected language models are available.

My thought is that when OSD is enabled, the user-provided languages are the last resort option. If one of the user-provided languages happens to also be automatically detected, then we should use it of course; otherwise, I think this component should just skip the missing OSD-detected language and move on to the next one that meets our thresholds, if one exists. If we exhausted the list and haven't found any of those models, then use the user-provided models.

Is there a reason you think this might not be a good approach? Generally, my thought is that using the user-provided languages may produce garbage, and it's better to always use the OSD-detected languages if we have the option. There's no need to mix garbage tracks in with good tracks if we can avoid it.

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 719 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/126#-LsIH8TgD1NJifTIzqzL:-LuImZzK-En_Aptc2hUr:b-z1gms5) ([raw file](https://github.com/openmpf/openmpf-components/blob/59a10f92993ebedbbc331943549644b35bed6b15/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L719)):*

This has not been addressed yet and will be explored in a future PR after the Docker work. 
</blockquote></details>

I created this a wishlist item. Don't worry about it for now: https://github.com/openmpf/openmpf/issues/1008

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 846 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/126#-Lv6mX409kUdlRObK_ca:-Lv6mX41CgdyPS30CU1Q:baxizqr) ([raw file](https://github.com/openmpf/openmpf-components/blob/64f517b1a1066133ec9aebec1d810b0a4b1abba4/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L846)):*
> ```cpp
>             }
>             if(!process_tesseract_lang_model(inputs, &ocr_var)) {
>                     return false;
> ```

I ran this test:

```
sample_tesseract_ocr_detector -i /home/mpf/git/openmpf-projects/openmpf-components/cpp/TesseractOCRTextDetection/test/data/eng-bul.png tur
```

Note that "tur" does not exist. I set some breakpoints and saw that when this `return false` line is called, `ocr_var.job_status` is MPF_COULD_NOT_OPEN_DATAFILE, as expected. However, consider the code one step up the call stack:

```
        if ( !get_tesseract_detections(&ocr_job_inputs, &image_results)) {
            LOG4CXX_ERROR(hw_logger_, "[" + job.job_name + "] Could not run tesseract!");
            return *image_results.job_status;
        }
```

Here, `image_results.job_status` is MPF_DETECTION_SUCCESS. Thus, the error status is lost.

Looking at your code

```
            ocr_var.job_status = *results->job_status;
            ocr_var.lang = lang;

            *results->job_status = ocr_var.job_status;
```

This is confusing. The first line is setting the job status to the current results job status. The last line is then setting the results job status to the same value it already has.

I think what you're trying to do in the last line is:

```
results->job_status = &ocr_var.job_status;
```

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 846 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/126#-Lv6mX409kUdlRObK_ca:-Lv7BbXx-BuvR7Ewd-pC:brxp6z2) ([raw file](https://github.com/openmpf/openmpf-components/blob/64f517b1a1066133ec9aebec1d810b0a4b1abba4/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L846)):*

I ran this test:

```
sample_tesseract_ocr_detector -i /home/mpf/git/openmpf-projects/openmpf-components/cpp/TesseractOCRTextDetection/test/data/eng-bul.png tur
```

Note that "tur" does not exist. I set some breakpoints and saw that when this `return false` line is called, `ocr_var.job_status` is MPF_COULD_NOT_OPEN_DATAFILE, as expected. However, consider the code one step up the call stack:

```
        if ( !get_tesseract_detections(&ocr_job_inputs, &image_results)) {
            LOG4CXX_ERROR(hw_logger_, "[" + job.job_name + "] Could not run tesseract!");
            return *image_results.job_status;
        }
```

Here, `image_results.job_status` is MPF_DETECTION_SUCCESS. Thus, the error status is lost.

Looking at your code

```
            ocr_var.job_status = *results->job_status;
            ocr_var.lang = lang;

            *results->job_status = ocr_var.job_status;
```

This is confusing. The first line is setting the job status to the current results job status. The last line is then setting the results job status to the same value it already has.

I think what you're trying to do in the last line is:

```
results->job_status = &ocr_var.job_status;
```
</blockquote></details>

Actually, don't do:

```
results->job_status = &ocr_var.job_status;
``` 

You do not want to set the reference to a local variable inside `get_tesseract_detections` because strange things will happen when trying to access it once you've left that function.

Instead consider changing `OCR_results.job_status` to a pointer:

```
            struct OCR_results {
                std::string text_result;
                std::string lang;
                MPFDetectionError *job_status;
                double confidence;
                bool parallel_processing;
                std::set<std::string> missing_languages;
            };
```

Which is how you have it in `Image_results`.

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 846 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/126#-Lv6mX409kUdlRObK_ca:-Lv7FVxT8jsnljgqRbsV:b9nmo2n) ([raw file](https://github.com/openmpf/openmpf-components/blob/64f517b1a1066133ec9aebec1d810b0a4b1abba4/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L846)):*

Actually, don't do:

```
results->job_status = &ocr_var.job_status;
``` 

You do not want to set the reference to a local variable inside `get_tesseract_detections` because strange things will happen when trying to access it once you've left that function.

Instead consider changing `OCR_results.job_status` to a pointer:

```
            struct OCR_results {
                std::string text_result;
                std::string lang;
                MPFDetectionError *job_status;
                double confidence;
                bool parallel_processing;
                std::set<std::string> missing_languages;
            };
```

Which is how you have it in `Image_results`.
</blockquote></details>

Once I changed it to a pointer, I was able to do:

```
            OCR_results ocr_var;
            ocr_var.job_status = results->job_status;
            ocr_var.lang = lang;

            if (inputs->ocr_fset->max_parallel_pdf_threads > 1 && inputs->process_pdf) {
                ocr_var.parallel_processing = true;
            } else {
                ocr_var.parallel_processing = false;

            }
```

Of course, I had to change a number of other lines to use the pointer.

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 846 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/126#-Lv6mX409kUdlRObK_ca:-Lv7b8pG8122Ojsbm3m3:b9lhg7n) ([raw file](https://github.com/openmpf/openmpf-components/blob/64f517b1a1066133ec9aebec1d810b0a4b1abba4/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L846)):*

Once I changed it to a pointer, I was able to do:

```
            OCR_results ocr_var;
            ocr_var.job_status = results->job_status;
            ocr_var.lang = lang;

            if (inputs->ocr_fset->max_parallel_pdf_threads > 1 && inputs->process_pdf) {
                ocr_var.parallel_processing = true;
            } else {
                ocr_var.parallel_processing = false;

            }
```

Of course, I had to change a number of other lines to use the pointer.
</blockquote></details>

If they don't already exist, please create 4 unit tests that check the final job status when all of the user-specified models are missing (when not doing OSD):
- Process image without parallel threads
- Process image with parallel threads
- Process PDF without parallel threads
- Process PDF with parallel threads

Basically, I just want to test the job status propagation behavior for all of the possible flows.

jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 27 at r8](https://reviewable.io/reviews/openmpf/openmpf-components/126#-Ly_WHfIBhgxCDAkIB6o:-Ly_WHfJ7RqGCNlrrovd:b783vvb) ([raw file](https://github.com/openmpf/openmpf-components/blob/e78f49e5ac4464b8c299cb30bb14e88196c97b85/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L27)):*
> ```cpp
>  ******************************************************************************/
> 
> #include <map>
> ```

According to CLion, the following aren't used:

```
#include <boost/algorithm/string/regex.hpp>
#include <opencv2/highgui/highgui.hpp>
#include <log4cxx/logmanager.h>
#include <cctype>
#include <fstream>
```

PR #152, opened 2020-05-11
Ssd face

fpertl:


jrobble:
Assigning @hhuangMITRE to do an initial round of review.

jrobble:
___
*[cpp/OcvSsdFaceDetection/test/test_ocv_ssd_face_detection.cpp, line 424 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/152#-M7nR3yB69uioQD_cOKI:-M7zOtNR8yeeHRdtTVRw:b-g6zcco) ([raw file](https://github.com/openmpf/openmpf-components/blob/75a1cb88cf1147775eba77ff7008d94f3a785af1/cpp/OcvSsdFaceDetection/test/test_ocv_ssd_face_detection.cpp#L424)):*
<details><summary><i>Previously, fpertl wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Let's just leave this code commented out. I can see how it would be useful for a developer to debug this code, but also it doesn't really make sense to generate these files every time the test is run because most of a time a human is not going to look at them.

PR #155, opened 2020-05-15
Hf merge/tesseract exception handling

hhuangMITRE:
Merging Tesseract 4.1 updates from master. Resolving Tesseract image scaling limits and handling of nullptr returns from OCR runs.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/155)


jrobble:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 1164 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/155#-M7OYv3m1r3xhQ-uKpz3:-M7OhRmTAfXcLjGkOLUw:bussa7e) ([raw file](https://github.com/openmpf/openmpf-components/blob/2d6d7726523dc468c235fdb583322f91b620980e/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L1164)):*

@hhuangmitre , replicating the image four times down looks like new behavior. Did you add any other new behavior?

Does it make sense to just do:

```
        cv::Mat amplified_im = cv::repeat(imi_copy, 4, 4);
        if (!rescale_image(job, amplified_im, ocr_fset, job_placeholder)) {
```

Basically, copy across and down in the same operation and then rescale? That will save us one repeat call and one scale call.
</blockquote></details>

I see you reverted the replication behavior back. Is that because it's unlikely to have any benefit? As in, replicating the image 4 times is just as likely to be successful as replicating it 4x4=16 times?

PR #163, opened 2020-06-18
Fix tesseract memory leak and improve error handling.

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/163)

### DIFF
--- a/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp
+++ b/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp
@@ -67,6 +67,29 @@ typedef boost::multi_index::multi_index_container<string,
             >> unique_vector;
 
 
+class TempDirectory {
+public:
+    const std::string path;
+
+    explicit TempDirectory(const std::string &path)
+            : path(path) {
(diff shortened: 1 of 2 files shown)


brosenberg42:
___
*[cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp, line 892 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/163#-MA7eimdC-fAVjDHMkUk:-MA7eimdC-fAVjDHMkUl:b-qiudba) ([raw file](https://github.com/openmpf/openmpf-components/blob/e7bdda6d69a20968e99891f7ef4a8119355b35d7/cpp/TesseractOCRTextDetection/TesseractOCRTextDetection.cpp#L892)):*
> ```cpp
>                         results.job_status = ocr_results[i].job_status;
>                         active_threads.erase(i);
>                         // Terminate remaining threads early due to error.
> ```

This comment is inaccurate. It didn't terminate the threads early, it just waited for them all to exit. With my changes when an error occurs, it still waits for the threads to exit because future's destructor waits until the thread exits. Actually exiting early would greatly complicate the code, so it isn't worth it since errors should be rare.

PR #178, opened 2020-08-26
Enable OcvDnnDetection to annotate feed-forward detections. (#177)

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/178)


jrobble:
Will land on command line via fast-forward.

PR #182, opened 2020-09-04
Adding Azure Form Recognizer component.

hhuangMITRE:
Resolves openmpf/openmpf#1161.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)


brosenberg42 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 0 of 18 files reviewed, 9 unresolved discussions (waiting on @brosenberg42, @hhuangMITRE, and @jrobble)

___
*[python/AzureFormRecognizerDetection/sample_acs_ocr_text_detector.py, line 1 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmc0I42B6Gp7uPoIgw:-MGmc0I42B6Gp7uPoIgx:b-9q2ql3) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/sample_acs_ocr_text_detector.py#L1)):*
> ```Python
> #!/usr/bin/env python3
> ```

Please name this file appropriately. Currently "ocr" is in the file name.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 89 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGjCburCOTWAPws1E_8:-MGjCbusA-r9ZAG2Y3-l:b-me6lkd) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L89)):*
> ```Python
>         self._acs_url = self.get_acs_url(job_properties)
>         self._is_image = is_image
>         self._merge_lines = json.loads(job_properties.get("MERGE_LINES", "true").lower())
> ```

Parts of this file use `'` and others use `"`. Please use `'` except for docstrings.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 102 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGma91YDh4nr4MwdtAL:-MGma91YDh4nr4MwdtAM:bfiavc9) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L102)):*
> ```Python
>         self._text_tagger = TextTagger(job_properties)
> 
>     def create_detection(self, detection_properties):
> ```

Please use a leading underscore for methods that aren't used outside of their own class.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 121 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGjDJEK4jHxi7AYaxUg:-MGjDJEK4jHxi7AYaxUh:bmo693j) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L121)):*
> ```Python
>             for line in page["lines"]:
>                 if not self._merge_lines:
>                     detection_properties = dict(OUTPUT_TYPE = "LINE",
> ```

pep8 is the common Python style guide. It says your are not supposed to have a space around the `=` when using named arguments.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 172 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGjEUAkBrzK7U28QMTt:-MGjDnm4BxSYwAf5sq6R:b-u7j1yd) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L172)):*
> ```Python
>         """
>         :param frames: An iterable of frames in a numpy.ndarray with shape (h, w, 3)
>         :return: An iterable of (frame_number, mpf.ImageLocation)
> ```

This comment is not accurate. You are returning an iterable of `mpf.ImageLocation`.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 176 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmdBl1CrZH9_LIFA02:-MGmdBl1CrZH9_LIFA03:b2kalla) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L176)):*
> ```Python
>         encoder = FrameEncoder()
> 
>         for idx, frame in enumerate(frames):
> ```

Since this component does not support video, the function should be modified to accept a single frame. Then you won't need this loop. You will still need the body, it just doesn't need to be in a loop. You will also need to change 
`detections = JobRunner(image_job.job_properties, True).get_image_detections((image,))` 
to 
`detections = JobRunner(image_job.job_properties, True).get_image_detections(image)`
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 193 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmxz6nAUfn1xTbRQDm:-MGmxz6o052Tmpe2UcR9:b-m5be1f) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L193)):*
> ```Python
>             response = urllib.request.urlopen(request)
>             results_url = response.headers["operation-location"]
>             result_request = urllib.request.Request(results_url, None, self._acs_headers)
> ```

When trying to run jobs through the command line, it kept failing with: `KeyError: 'analyzeResult'`. I printed the JSON and it contained: `{'status': 'running', 'createdDateTime': '2020-09-09T13:00:14Z', 'lastUpdatedDateTime': '2020-09-09T13:00:14Z'}`. I think you need to poll until `status` is either `succeeded` or `failed`. I think you will need to do a little bit of testing to determine a good interval to sleep between each poll. ACS OCR doesn't need to poll because that REST API in synchronous. 

It would also be good to log when you post the job and the result of each poll attempt. 
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 338 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGn3dzG1CX2AaFsgChX:-MGn3Wh-6zWXnPRrFS-L:b-3m7n4c) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L338)):*
> ```Python
> 
>     # These constraints are from Azure Cognitive Services
>     # (https://westus.dev.cognitive.microsoft.com/docs/services/56f91f2d778daf23d8ec6739/operations/56f91f2e778daf14a499e1fc)
> ```

The link here goes to the OCR REST API. The input constraints vary from service to service. The constraints from, https://westus2.dev.cognitive.microsoft.com/docs/services/form-recognizer-api-v2/operations/AnalyzeLayoutAsync , are:
```
MIN_DIMENSION_LENGTH = 50  # pixels
MAX_DIMENSION_LENGTH = 10000  # pixels
MAX_FILE_SIZE = 20 * 1024 * 1024  # bytes
```
The form API doesn't seem to have a maximum number of pixels like OCR does.
___
*[python/AzureFormRecognizerDetection/tests/test_acs.py, line 48 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmcifD6DKJON7TFAoL:-MGmcifD6DKJON7TFAoM:blu96m0) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/tests/test_acs.py#L48)):*
> ```Python
> 
> 
> class TestAcs(unittest.TestCase):
> ```

Please add tests for pdfs and documents with actual forms or tables in them.

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 0 of 20 files reviewed, 9 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 89 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGjCburCOTWAPws1E_8:-MHoQxOW6wgZm8ZTS92j:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L89)):*

Parts of this file use `'` and others use `"`. Please use `'` except for docstrings.
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 102 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGma91YDh4nr4MwdtAL:-MHoR-pgAoKyDFYLQr4n:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L102)):*

Please use a leading underscore for methods that aren't used outside of their own class.
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 121 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGjDJEK4jHxi7AYaxUg:-MHoR0Z96Hyvzg6SDw11:bj3eodx) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L121)):*

pep8 is the common Python style guide. It says your are not supposed to have a space around the `=` when using named arguments.
</blockquote></details>

Fixed, thanks!
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 172 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGjEUAkBrzK7U28QMTt:-MHoTcCa4Ovr76kUK6Pm:b-jtcuzo) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L172)):*

This comment is not accurate. You are returning an iterable of `mpf.ImageLocation`.
</blockquote></details>

Thanks for spotting that, I've corrected the docstring.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 176 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmdBl1CrZH9_LIFA02:-MHoTtDfA4RxhR7VZhlp:b-pegv1l) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L176)):*

Since this component does not support video, the function should be modified to accept a single frame. Then you won't need this loop. You will still need the body, it just doesn't need to be in a loop. You will also need to change 
`detections = JobRunner(image_job.job_properties, True).get_image_detections((image,))` 
to 
`detections = JobRunner(image_job.job_properties, True).get_image_detections(image)`
</blockquote></details>

That makes sense, I've updated the JobRunner to process single frame inputs and removed the loop.

Unrelated, but I've also removed the Text Tagger code as we've determined that this is not a priority/need right now. Keyword tagger can also process "TEXT" results from this component.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 193 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmxz6nAUfn1xTbRQDm:-MHoUGru5AkEO_uecrl_:b-50otho) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L193)):*

When trying to run jobs through the command line, it kept failing with: `KeyError: 'analyzeResult'`. I printed the JSON and it contained: `{'status': 'running', 'createdDateTime': '2020-09-09T13:00:14Z', 'lastUpdatedDateTime': '2020-09-09T13:00:14Z'}`. I think you need to poll until `status` is either `succeeded` or `failed`. I think you will need to do a little bit of testing to determine a good interval to sleep between each poll. ACS OCR doesn't need to poll because that REST API in synchronous. 

It would also be good to log when you post the job and the result of each poll attempt. 
</blockquote></details>

Sounds good,  I've ran more tests, and referred to the official REST API examples for a good polling interval and updated the job run to sleep 5 seconds between polls. Logging has been updated to post each poll attempt as well.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 338 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGn3dzG1CX2AaFsgChX:-MHoVK3_8cCvtPl8dtFv:b-7gjmr9) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L338)):*

The link here goes to the OCR REST API. The input constraints vary from service to service. The constraints from, https://westus2.dev.cognitive.microsoft.com/docs/services/form-recognizer-api-v2/operations/AnalyzeLayoutAsync , are:
```
MIN_DIMENSION_LENGTH = 50  # pixels
MAX_DIMENSION_LENGTH = 10000  # pixels
MAX_FILE_SIZE = 20 * 1024 * 1024  # bytes
```
The form API doesn't seem to have a maximum number of pixels like OCR does.
</blockquote></details>

Oh I see, I've updated the dimension/file size limits following ACS form recognizer limits (thanks for highlighting the altered values) and also removed the MAX_PIXELS limit + checks since they're not present for this service. The unit tests for the constraints have been updated as well.
___
*[python/AzureFormRecognizerDetection/tests/test_acs.py, line 48 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmcifD6DKJON7TFAoL:-MHoWjct5BysLtnfiBkG:bfciv9n) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/tests/test_acs.py#L48)):*

Please add tests for pdfs and documents with actual forms or tables in them. 
</blockquote></details>

Done. Also updated the tests for the image size limits. 
___
*[python/AzureFormRecognizerDetection/sample_acs_ocr_text_detector.py, line 1 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MGmc0I42B6Gp7uPoIgw:-MHoQh6OEnk8mlbzJD--:b-1rym30) ([raw file](https://github.com/openmpf/openmpf-components/blob/25437d344915b4e2808a767c914eac6b6d510583/python/AzureFormRecognizerDetection/sample_acs_ocr_text_detector.py#L1)):*

Please name this file appropriately. Currently "ocr" is in the file name.
</blockquote></details>

Thanks for catching that. I've changed the file name to "sample_acs_form_recognizer.py".

brosenberg42:
___
*[python/AzureFormRecognizerDetection/tests/test_acs.py, line 44 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqwinU2MJ7X9LLAvmz:-MHqwinU2MJ7X9LLAvn-:b-iz2dqc) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/tests/test_acs.py#L44)):*
> ```Python
> from acs_form_recognizer_component import AcsFormRecognizerComponent, FrameEncoder
> 
> sys.path.insert(0, os.path.join(os.path.dirname(__file__), '../acs_form_recognizer_component'))
> ```

The tests fail in the Docker build because you need to modify `sys.path` before importing anything form `acs_form_recognizer_component`

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 0 of 20 files reviewed, 10 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[python/AzureFormRecognizerDetection/tests/test_acs.py, line 44 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqwinU2MJ7X9LLAvmz:-MHr4J4n-t-Urk-OB4Mt:beygvvm) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/tests/test_acs.py#L44)):*

The tests fail in the Docker build because you need to modify `sys.path` before importing anything form `acs_form_recognizer_component`
</blockquote></details>

Fixed, thanks for spotting that.

brosenberg42 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 0 of 20 files reviewed, 13 unresolved discussions (waiting on @hhuangMITRE and @jrobble)

___
*[python/AzureFormRecognizerDetection/setup.py, line 31 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqsOIe7t93Snnn6Vwx:-MHqsOIe7t93Snnn6Vwy:b-4nn9tb) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/setup.py#L31)):*
> ```Python
> 
> setuptools.setup(
>     name='AzureFormRecognizerDetection',
> ```

I was looking through my emails about this and we intended for the component to be named `AzureFormDetection`. Please update the directory name, file names, and class names to use this name.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 73 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqaVtVEwGlq4tubWH7:-MHqaVtVEwGlq4tubWH8:b4i4uzl) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L73)):*
> ```Python
>             detections = JobRunner(image_job.job_name, image_job.job_properties, True).get_image_detections(image)
>             for detection in detections:
>                 # We don't need to anything with the index for a image jobs.
> ```

This comment is no longer relevant.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 89 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHrBx9qAwxENpNqKXtR:-MHrBx9qAwxENpNqKXtS:b7j7zi) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L89)):*
> ```Python
>         self._acs_url = self.get_acs_url(job_properties)
>         self._is_image = is_image
>         self._merge_lines = json.loads(job_properties.get('MERGE_LINES', 'true').lower())
> ```

Please change this line to: `self._merge_lines = mpf_util.get_property(job_properties, 'MERGE_LINES', True)`
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 118 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqa2NW8gDzCniHh_m9:-MHqa2NW8gDzCniHh_mA:b-olep32) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L118)):*
> ```Python
>                                           resize_scale_factor,
>                                           self._merge_lines).process_form_results(form_results_json)
>         for detection in detections:
> ```

Since this is no longer in a nested loop, you can remove this loop and just `return detections` 
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 131 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqZrO2310Z9f-u7L0u:-MHqZrO2310Z9f-u7L0v:b-l9fpbh) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L131)):*
> ```Python
>     def _get_acs_result(self, result_request):
>         logger.info('[%s] Contacting ACS server for form results.', self._job_name)
>         max_attempts = 10
> ```

I don't think we necessarily need to set a limit on the maximum number of attempts. If we do, we should make it configurable as a job property.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 140 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqjKt49BSGcpBbQjv4:-MHqjKt49BSGcpBbQjv5:b1zdvtl) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L140)):*
> ```Python
>                 json_results = json.load(results)
>                 if results.status != 200:
>                     logger.warning("GET Layout results failed:\n{}".format(json_results))
> ```

Logging the error here isn't necessary because it will be logged again by your top-level exception handlers. Please include the HTTP request body in the exception error message. Right now it is in the log message, but not the exception. The exception message will appear in the JSON output object.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 141 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqiVAB4FA5S2daNiO3:-MHqiVAB4FA5S2daNiO4:bdl6hsp) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L141)):*
> ```Python
>                 if results.status != 200:
>                     logger.warning("GET Layout results failed:\n{}".format(json_results))
>                     raise mpf.DetectionException('GET Request failed with HTTP status {}'.format(results.status))
> ```

Please set the error code to `mpf.DetectionError.DETECTION_FAILED` for http issues. Since you are using a newer version of the Python SDK, it can be done like this: 
`raise mpf.DetectionError.DETECTION_FAILED.exception('<msg>')`. 
You can also use the old way which is: 
`raise mpf.DetectionException('<msg>', mpf.DetectionError.DETECTION_FAILED)`
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 135 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHugHmLEH5moFBWS5XZ:-MHugHmLEH5moFBWS5X_:b-ei3wj) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L135)):*
> ```Python
>         wait_sec = 5
>         while attempt_num < max_attempts:
>             logger.info('[%s] Sending GET request.', self._job_name)
> ```

Please change this to:
```
logger.info('[%s] Attempting to retrieve layout analysis results from %s',
            self._job_name, result_request.get_full_url())
```
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 150 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHuhoNV4TDVnHgOd3Wq:-MHuhoNV4TDVnHgOd3Wr:b-s22l59) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L150)):*
> ```Python
>                     raise mpf.DetectionException('GET Request failed.')
>                 # Analysis still running. Wait and then retry GET request.
>                 logger.info('[%s] Form results not available yet. Recontacting server in [%d] seconds.', self._job_name,
> ```

All of the other logging calls it "Layout Analysis", not "Form results". Please change this to:
```
logger.info('[%s] Layout Analysis results not available yet. Recontacting server in %s seconds.',
            self._job_name, wait_sec)
```
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 162 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHugf5P57dTsf7-rfro:-MHugf5P57dTsf7-rfrp:bw0p6kn) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L162)):*
> ```Python
> 
>     def _post_to_acs(self, encoded_frame):
>         logger.info('[%s] Sending POST request to ACS server.', self._job_name)
> ```

Please change this to: `logger.info('[%s] Sending POST request to %s', self._job_name, self._acs_url)`
___
*[python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json, line 52 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHrAIHb7_msN1RIfFtw:-MHrAIHb7_msN1RIfFtx:bmbxi7j) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json#L52)):*
> ```JSON
>     {
>       "name": "AZURE FORM RECOGNIZER DETECTION ACTION",
>       "description": "Uses Azure Cognitive Services to perform optical character recognition.",
> ```

These descriptions still refer to "optical character recognition". Please update the descriptions.
___
*[python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json, line 68 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHr9INV2dsDreLYprcZ:-MHr9INV2dsDreLYprc_:bkxvfm2) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json#L68)):*
> ```JSON
>   "pipelines": [
>     {
>       "name": "AZURE FORM RECOGNIZER DETECTION PIPELINE",
> ```

Please add a `AZURE FORM DETECTION (WITH MARKUP) PIPELINE` pipeline. It will almost identical to this one. The only difference is that `"tasks"` will be:
```
"tasks": [
  "AZURE FORM DETECTION TASK",
  "OCV GENERIC MARKUP TASK"
]
```
___
*[python/AzureFormRecognizerDetection/tests/test_acs.py, line 271 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHrHDVC3lSq88lQUkA-:-MHrHDVD1qJL9GANDNs9:b32i7qw) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/tests/test_acs.py#L271)):*
> ```Python
>             raise Exception(msg)
> 
>         if not post_body.startswith(b'\x89PNG\x0d\x0a\x1a\x0a') and not post_body.startswith(b'%PDF'):
> ```

The `not post_body.startswith(b'%PDF')` part of the conditional is never actually checked because you only call this method when `self.headers['Content-Type'] == 'application/octet-stream'`
You could change this method to `_validate_body`, then exit early when `post_body.startswith(b'%PDF')` is `True` since the frame size checks only apply to images.

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 0 of 22 files reviewed, 13 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[python/AzureFormRecognizerDetection/setup.py, line 31 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqsOIe7t93Snnn6Vwx:-MIRY7Ji-FAk37X0LDuk:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/setup.py#L31)):*

I was looking through my emails about this and we intended for the component to be named `AzureFormDetection`. Please update the directory name, file names, and class names to use this name.
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 73 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqaVtVEwGlq4tubWH7:-MIRY9rZ5mcsxM_WrcZt:b-pctro8) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L73)):*

This comment is no longer relevant.
</blockquote></details>

Remove, thanks for spotting that.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 89 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHrBx9qAwxENpNqKXtR:-MIRYFdI8o2yCWYJ-FRm:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L89)):*

Please change this line to: `self._merge_lines = mpf_util.get_property(job_properties, 'MERGE_LINES', True)`
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 118 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqa2NW8gDzCniHh_m9:-MIRYTX47enOJSALdQBQ:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L118)):*

Since this is no longer in a nested loop, you can remove this loop and just `return detections` 
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 131 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqZrO2310Z9f-u7L0u:-MIRYUUy56giT_rLbxuf:bl2yl61) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L131)):*

I don't think we necessarily need to set a limit on the maximum number of attempts. If we do, we should make it configurable as a job property.
</blockquote></details>

Done. I've updated the retries to continue without limits unless the user wishes to set a max number of attempts. 
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 140 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqjKt49BSGcpBbQjv4:-MIRYk_W3ant0oeoG4KO:b-u4m676) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L140)):*

Logging the error here isn't necessary because it will be logged again by your top-level exception handlers. Please include the HTTP request body in the exception error message. Right now it is in the log message, but not the exception. The exception message will appear in the JSON output object.
</blockquote></details>

Understood, thanks! Since the messages will be logged during exception handling I've removed the manual error log entries here, and moved the HTTP request body into the exception error message.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 141 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHqiVAB4FA5S2daNiO3:-MIRZSOaDz_DyaihNI5F:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L141)):*

Please set the error code to `mpf.DetectionError.DETECTION_FAILED` for http issues. Since you are using a newer version of the Python SDK, it can be done like this: 
`raise mpf.DetectionError.DETECTION_FAILED.exception('<msg>')`. 
You can also use the old way which is: 
`raise mpf.DetectionException('<msg>', mpf.DetectionError.DETECTION_FAILED)`
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 135 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHugHmLEH5moFBWS5XZ:-MIRYjPy9Ks8R5wSMd92:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L135)):*

Please change this to:
```
logger.info('[%s] Attempting to retrieve layout analysis results from %s',
            self._job_name, result_request.get_full_url())
```
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 150 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHuhoNV4TDVnHgOd3Wq:-MIRZeCK4JbKJy15kl5L:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L150)):*

All of the other logging calls it "Layout Analysis", not "Form results". Please change this to:
```
logger.info('[%s] Layout Analysis results not available yet. Recontacting server in %s seconds.',
            self._job_name, wait_sec)
```
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py, line 162 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHugf5P57dTsf7-rfro:-MIRZgQO4tVp_V9Tlxjt:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/acs_form_recognizer_component/acs_form_recognizer_component.py#L162)):*

Please change this to: `logger.info('[%s] Sending POST request to %s', self._job_name, self._acs_url)`
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json, line 52 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHrAIHb7_msN1RIfFtw:-MIRZhWHAQcMRfObWbFO:b-4xpm6f) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json#L52)):*

These descriptions still refer to "optical character recognition". Please update the descriptions.
</blockquote></details>

Fixed, thanks for catching that.
___
*[python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json, line 68 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHr9INV2dsDreLYprcZ:-MIRZo9o7MCeF_wSMplE:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/545e677525205863df1c962beb0e369e9285440d/python/AzureFormRecognizerDetection/plugin-files/descriptor/descriptor.json#L68)):*

Please add a `AZURE FORM DETECTION (WITH MARKUP) PIPELINE` pipeline. It will almost identical to this one. The only difference is that `"tasks"` will be:
```
"tasks": [
  "AZURE FORM DETECTION TASK",
  "OCV GENERIC MARKUP TASK"
]
```
</blockquote></details>

Done.
___
*[python/AzureFormRecognizerDetection/tests/test_acs.py, line 271 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MHrHDVC3lSq88lQUkA-:-MIRZpm80UrOocwc1uUJ:b7hrxkc) ([raw file](https://github.com/openmpf/openmpf-components/blob/1dfd23732c5808a0fa44fc81e2c8a429004cd56c/python/AzureFormRecognizerDetection/tests/test_acs.py#L271)):*

The `not post_body.startswith(b'%PDF')` part of the conditional is never actually checked because you only call this method when `self.headers['Content-Type'] == 'application/octet-stream'`
You could change this method to `_validate_body`, then exit early when `post_body.startswith(b'%PDF')` is `True` since the frame size checks only apply to images.
</blockquote></details>

Sounds good. I've updated the function to `_validate_body` following the suggestion.
I've moved the `self.headers` check over here so that we check for PDF files when the `Content-Type` is set to `application/pdf`, and PNG files otherwise.

brosenberg42 (changes requested):
> Reviewed 21 of 23 files at r4.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 21 of 22 files reviewed, 3 unresolved discussions (waiting on @hhuangMITRE and @jrobble)

___
*[python/AzureFormDetection/README.md, line 39 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MITyizW8HC1jdLb5Dme:-MITyizW8HC1jdLb5Dmf:b-21ccv4) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/README.md#L39)):*
> ```Markdown
> - `LANGUAGE`:  When provided, disables automatic language detection and processes document based on provided language.
>    The language code must be in case-sensitive BCP-47.  Currently, only English (`en`), Dutch (‘nl’),
>    French (‘fr’), German (‘de’), Italian (‘it’), Portuguese (‘pt`), simplified Chinese (`zh-Hans`) and Spanish (`es`)
> ```

If you go to https://github.com/openmpf/openmpf-components/blob/azure-form-recognizer/python/AzureFormDetection/README.md , you can see the rendered version of this markup. This part is messed up because pt is using an apostrophe (') on the left side and a back-tick (\`) on the other. For all of the abbreviations, please use the same quoting scheme. You can use back-ticks (\`), quotes ("), or nothing at all. 
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 188 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MITl-Xe4vYDlK9iZoyw:-MITl-Xf4aKw9o0M1Nfr:b-6paeef) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L188)):*
> ```Python
>         query_string = urllib.parse.urlencode(query_dict, doseq=True)
>         if language:
>             query_dict['language'] = language
> ```

You are modifying query_dict after it has already been encoded, so it has no effect.
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 497 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIUPCFo94SiWmlunQdT:-MIUPCFo94SiWmlunQdU:bdeeyzy) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L497)):*
> ```Python
>         if 'documentResults' in form_results_json['analyzeResult']:
>             doc_result_index = 0
>             for doc_result in form_results_json['analyzeResult']['documentResults']:
> ```

Shouldn't we pull out the key value pairs here? Right now DOCUMENT_JSON_FIELDS is just the raw JSON from Azure. I ran `python sample_acs_form_detector.py 'https://example.com/formrecognizer/v2.1-preview.1/prebuilt/receipt/analyze' <subscription key> tests/data/regular-forms/contoso-invoice.png` and the
DOCUMENT_JSON_FIELDS property contained: `{'ReceiptType': {'type': 'string', 'valueString': 'Itemized', 'confidence': 0.979}, 'TransactionDate': {'type': 'date', 'valueDate': '2020-09-10', 'text': '9/10/2020', 'boundingBox': [1050, 114, 1132, 114, 1132, 136, 1050, 135], 'page': 1, 'confidence': 0.844, 'elements': ['#/readResults/0/lines/2/words/1']}, 'MerchantName': {'type': 'string', 'valueString': 'Sold To Wingtip Toys', 'text': 'Sold To Wingtip Toys', 'boundingBox': [191.2, 188, 385.2, 189.7, 385, 210.3, 191, 208.6], 'page': 1, 'confidence': 0.516, 'elements': ['#/readResults/0/lines/4/words/0', '#/readResults/0/lines/4/words/1', '#/readResults/0/lines/4/words/2', '#/readResults/0/lines/4/words/3']}, 'MerchantAddress': {'type': 'string', 'valueString': '554 Magnolia Way Walnut, WY, 98432', 'text': '554 Magnolia Way Walnut, WY, 98432', 'boundingBox': [285.6, 238.1, 431, 240, 430.4, 282.2, 285, 280.2], 'page': 1, 'confidence': 0.984, 'elements': ['#/readResults/0/lines/5/words/0', '#/readResults/0/lines/5/words/1', '#/readResults/0/lines/5/words/2', '#/readResults/0/lines/6/words/0', '#/readResults/0/lines/6/words/1', '#/readResults/0/lines/6/words/2']}, 'Subtotal': {'type': 'number', 'valueNumber': 10375, 'text': '$10,375', 'boundingBox': [1012, 868, 1087, 868, 1088, 891, 1012, 890], 'page': 1, 'confidence': 0.99, 'elements': ['#/readResults/0/lines/29/words/0']}, 'Total': {'type': 'number', 'valueNumber': 10686.25, 'text': '$10,686.25', 'boundingBox': [997, 945, 1105, 946, 1104, 968, 997, 968], 'page': 1, 'confidence': 0.947, 'elements': ['#/readResults/0/lines/33/words/0']}}`

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 19 of 22 files reviewed, 3 unresolved discussions (waiting on @brosenberg42, @hhuangMITRE, and @jrobble)

___
*[python/AzureFormDetection/README.md, line 39 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MITyizW8HC1jdLb5Dme:-MIUrbqI4GPUJGvuC0Yo:bwfhmfc) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/README.md#L39)):*

If you go to https://github.com/openmpf/openmpf-components/blob/azure-form-recognizer/python/AzureFormDetection/README.md , you can see the rendered version of this markup. This part is messed up because pt is using an apostrophe (') on the left side and a back-tick (\`) on the other. For all of the abbreviations, please use the same quoting scheme. You can use back-ticks (\`), quotes ("), or nothing at all. 
</blockquote></details>

Thanks, updated to backticks. Also, corrected the URL description as well. Will update again after making changes to document results.
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 188 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MITl-Xe4vYDlK9iZoyw:-MIUsix67S1nwTQhuBsm:b-s7skbj) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L188)):*

You are modifying query_dict after it has already been encoded, so it has no effect.
</blockquote></details>

Fixed, thanks for spotting that. Adding a check to make sure this is properly passed in similar to the one for AzureOCRTextDetection.
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 497 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIUPCFo94SiWmlunQdT:-MIUszyl05hqaqVYUGMZ:b-dl7hv2) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L497)):*

Shouldn't we pull out the key value pairs here? Right now DOCUMENT_JSON_FIELDS is just the raw JSON from Azure. I ran `python sample_acs_form_detector.py 'https://example.com/formrecognizer/v2.1-preview.1/prebuilt/receipt/analyze' <subscription key> tests/data/regular-forms/contoso-invoice.png` and the
DOCUMENT_JSON_FIELDS property contained: `{'ReceiptType': {'type': 'string', 'valueString': 'Itemized', 'confidence': 0.979}, 'TransactionDate': {'type': 'date', 'valueDate': '2020-09-10', 'text': '9/10/2020', 'boundingBox': [1050, 114, 1132, 114, 1132, 136, 1050, 135], 'page': 1, 'confidence': 0.844, 'elements': ['#/readResults/0/lines/2/words/1']}, 'MerchantName': {'type': 'string', 'valueString': 'Sold To Wingtip Toys', 'text': 'Sold To Wingtip Toys', 'boundingBox': [191.2, 188, 385.2, 189.7, 385, 210.3, 191, 208.6], 'page': 1, 'confidence': 0.516, 'elements': ['#/readResults/0/lines/4/words/0', '#/readResults/0/lines/4/words/1', '#/readResults/0/lines/4/words/2', '#/readResults/0/lines/4/words/3']}, 'MerchantAddress': {'type': 'string', 'valueString': '554 Magnolia Way Walnut, WY, 98432', 'text': '554 Magnolia Way Walnut, WY, 98432', 'boundingBox': [285.6, 238.1, 431, 240, 430.4, 282.2, 285, 280.2], 'page': 1, 'confidence': 0.984, 'elements': ['#/readResults/0/lines/5/words/0', '#/readResults/0/lines/5/words/1', '#/readResults/0/lines/5/words/2', '#/readResults/0/lines/6/words/0', '#/readResults/0/lines/6/words/1', '#/readResults/0/lines/6/words/2']}, 'Subtotal': {'type': 'number', 'valueNumber': 10375, 'text': '$10,375', 'boundingBox': [1012, 868, 1087, 868, 1088, 891, 1012, 890], 'page': 1, 'confidence': 0.99, 'elements': ['#/readResults/0/lines/29/words/0']}, 'Total': {'type': 'number', 'valueNumber': 10686.25, 'text': '$10,686.25', 'boundingBox': [997, 945, 1105, 946, 1104, 968, 997, 968], 'page': 1, 'confidence': 0.947, 'elements': ['#/readResults/0/lines/33/words/0']}}`
</blockquote></details>

Yes, I was wondering about that as well. For all string based keys, I can pull them in as key-value pairs without any issue. However, for keys associated with JSON arrays and JSON objects, I wanted to leave those in a format where users could still parse them out. So I'm wondering if this works:

Convert all keys with text-based values into key value pair outputs. For keys paired with JSON arrays and JSON objects, store them as JSON outputs in separate tracks (i.e. `DOC_ARRAY`) and (`DOC_OBJECT`) paired with their key ('DOC_KEY') value and document track index. Similar to how we're separating CSV-formatted tables from text lines and key-value pairs. 

If that's ok I'll make the changes and update the PR later today (occupied on unrelated task).

brosenberg42:
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 497 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIUPCFo94SiWmlunQdT:-MIZvh8c09bhdclOwkhr:b433ade) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L497)):*

Yes, I was wondering about that as well. For all string based keys, I can pull them in as key-value pairs without any issue. However, for keys associated with JSON arrays and JSON objects, I wanted to leave those in a format where users could still parse them out. So I'm wondering if this works:

Convert all keys with text-based values into key value pair outputs. For keys paired with JSON arrays and JSON objects, store them as JSON outputs in separate tracks (i.e. `DOC_ARRAY`) and (`DOC_OBJECT`) paired with their key ('DOC_KEY') value and document track index. Similar to how we're separating CSV-formatted tables from text lines and key-value pairs. 

If that's ok I'll make the changes and update the PR later today (occupied on unrelated task). 

</blockquote></details>

I'll have to discuss with @jrobble tomorrow to decide how to handle it.

jrobble:
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 497 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIUPCFo94SiWmlunQdT:-MIaiqpa9uy27_nrnG13:b-asj7o6) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L497)):*

I'll have to discuss with @jrobble tomorrow to decide how to handle it.
</blockquote></details>

Looking at @brosenberg42 's output above, I see that certain fields were identified: `TransactionDate`, `MerchantName`, `MerchantAddress`, ... I think these fields are different from what we've been calling key-value pairs in that these "keys" don't actually appear on the form at all (e.g. no where on the form does it say "TransactionDate"). Rather, they are names of fields hard-coded into the model.

Please send me an example of a more complicated output with JSON arrays / JSON objects. Can a string-based field be embedded in the results structure?  I don't want to tackle the problem of trying to recursively pull out sub-fields as key-value pairs unless we're sure that's what the end users want. I think we should give them a first cut of this component first to try out. One reason is that pulling embedded fields out results in losing the hierarchical structure, which may be desirable.

In general, I'm leaning towards keeping the structure of the output the same so it matches the structure the custom model was trained to detect, but I can be swayed.

One thing that might be worth considering is dropping the `boundingBox` and `elements` fields if we're not using them. They just add bloat.

jrobble:
___
*[python/AzureFormDetection/tests/test_acs.py, line 144 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIanoqV0cf5a68hjqGe:-MIanoqV0cf5a68hjqGf:b-nti1jw) ([raw file](https://github.com/openmpf/openmpf-components/blob/21b31fb7dd48d14533e5f132b465ff12efd90058/python/AzureFormDetection/tests/test_acs.py#L144)):*
> ```Python
>         table_detection = detections[1]
>         self.assertEqual('KEY_VALUE_PAIRS', table_detection.detection_properties['OUTPUT_TYPE'])
>         self.assertTrue('Date[:]:9/10/2020' in
> ```

Just looking at the code without running anything, I'm confused. I think this test is for `form-recognizer-demo-sample1.pdf`, but the 9/10/2020 date is on `contoso-invoice.png`. I don't see that field anywhere on`form-recognizer-demo-sample1.pdf`.

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 18 of 23 files reviewed, 4 unresolved discussions (waiting on @brosenberg42, @hhuangMITRE, and @jrobble)

___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 188 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MITl-Xe4vYDlK9iZoyw:-MIeI3QD95gXsJucc5-o:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L188)):*

Fixed, thanks for spotting that. Adding a check to make sure this is properly passed in similar to the one for AzureOCRTextDetection.
</blockquote></details>

Done.
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 497 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIUPCFo94SiWmlunQdT:-MIeJG4_EuUpIGf-MkP7:b-ldhiwl) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L497)):*

Looking at @brosenberg42 's output above, I see that certain fields were identified: `TransactionDate`, `MerchantName`, `MerchantAddress`, ... I think these fields are different from what we've been calling key-value pairs in that these "keys" don't actually appear on the form at all (e.g. no where on the form does it say "TransactionDate"). Rather, they are names of fields hard-coded into the model.

Please send me an example of a more complicated output with JSON arrays / JSON objects. Can a string-based field be embedded in the results structure?  I don't want to tackle the problem of trying to recursively pull out sub-fields as key-value pairs unless we're sure that's what the end users want. I think we should give them a first cut of this component first to try out. One reason is that pulling embedded fields out results in losing the hierarchical structure, which may be desirable.

In general, I'm leaning towards keeping the structure of the output the same so it matches the structure the custom model was trained to detect, but I can be swayed.

One thing that might be worth considering is dropping the `boundingBox` and `elements` fields if we're not using them. They just add bloat.


</blockquote></details>

@jrobble, I've sent over an example in Mattermost. I agree that overall the format of `documentResults` is quite different and much more flexible compared to other outputs  provided by ACS. 

Each individual entry in the `documentResults` JSON array holds one detected form. 
In each entry, there is a 'fields' dictionary pairing custom form keys with a particular value. 
Each field can be one of the following values:
```
valueString	String	Semantic value of "string" fields.
valueNumber	Number	Semantic value of "number" fields.
valuePhoneNumber	String	Semantic value of "phoneNumber" fields.
valueDate	String	Semantic value of "date" fields.
valueTime	String	Semantic value of "time" fields.
valueArray	[Object]	Semantic value of "array" fields.
valueObject	Object	Semantic value of "object" fields.
```

In short, `valueArray` and `valueObject` are the only fields that may contain complex, formatted outputs.

Since a user may want to process custom model outputs from ACS, I'm also leaning towards providing the `documentResults` as close to the original JSON formatting as possible. Right now, I'm converting each entry in `documentResults` into a separate output track containing the detected field results in the original JSON format, but we can also simplify results further later on if needed.

A possible simplification can involve splitting text-based pairings out from array and dictionary fields. Since most of the values are text (number, string, phone number, date, time) those key-value pairs can be placed into DOCUMENT_TEXT_KEY_VALUE_PAIRS, to allow for easier parsing of results. Arrays and objects could be output as separate detection tracks (DOCUMENT_ARRAY, DOCUMENT_OBJECT) and left in JSON string format. 

I also agree that since the "keys" here are not text found in the document, they should not be included with the text-based KEY_VALUE_PAIRS output. 
___
*[python/AzureFormDetection/tests/test_acs.py, line 144 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIanoqV0cf5a68hjqGe:-MIeI7xs-wvKmLfZ_7BF:b-4mozun) ([raw file](https://github.com/openmpf/openmpf-components/blob/21b31fb7dd48d14533e5f132b465ff12efd90058/python/AzureFormDetection/tests/test_acs.py#L144)):*

Just looking at the code without running anything, I'm confused. I think this test is for `form-recognizer-demo-sample1.pdf`, but the 9/10/2020 date is on `contoso-invoice.png`. I don't see that field anywhere on`form-recognizer-demo-sample1.pdf`.
</blockquote></details>

Thanks for catching that. Because KEY_VALUE_PAIRS only exist in custom models outputs, I found I had to use mock results (using examples from https://docs.microsoft.com/en-us/azure/cognitive-services/form-recognizer/quickstarts/python-train-extract?tabs=v2-1) to test for KEY_VALUE_PAIRS. 

Because the pdf and png jsons are real, I decided to move these mock key-value pairs into a separate mock JSON (based off of the contoso png file) and added a test to process the mock custom model output as well. Eventually, I'd like to generate a JSON from a real custom model.

brosenberg42 (changes requested):
> Reviewed 1 of 2 files at r5, 3 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 22 of 23 files reviewed, 6 unresolved discussions (waiting on @brosenberg42, @hhuangMITRE, and @jrobble)

___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 474 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIt0gLS2iUBIqhCskvF:-MIt0gLS2iUBIqhCskvG:b93you4) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L474)):*
> ```Python
> 
>             # Escape commas and colons with brackets as they are used as key-value delimiters.
>             key_str = kv_pair['key']['text'].replace(':', '[:]').replace(',', '[,]')
> ```

Since we are already using JSON in another field, I think it would be better to use JSON here so that consumers don't need to write their own custom parser for this field. You could just put the key-value pairs in a dictionary that is set up like: `{"key1": "value1", "key2": "value2"}` and then use `json.dumps(...)` to convert it to a JSON string.
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 485 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIt89_KCryEuBDEqziD:-MIt89_KCryEuBDEqziE:b-263lbz) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L485)):*
> ```Python
>         detections.append(self._create_detection(detection_properties, bounding_box))
> 
>     def _process_page_result_tables(self, page, page_num, detections):
> ```

Please add a test for processing `documentResults`. This function has no unit test coverage.
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 524 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIt-XgSDiRhAOwGTWc1:-MIt-XgTDt_PK3SHav9-:b9r9e7b) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L524)):*
> ```Python
>             # ACS bounding box and elements are then filtered out.
>             self._remove_elements(doc_result, bbox_list)
>             fields = str(doc_result['fields'])
> ```

This produces invalid JSON. It should be: `fields = json.dumps(doc_result['fields'])`
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 556 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIsqPSB7_m_6QAfuZsV:-MIsqPSC7JpT1j6ZkIYI:b5xiw21) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L556)):*
> ```Python
>         box = json_object.pop('boundingBox', None)
>         json_object.pop('elements', None)
>         print(box)
> ```

Please remove this print statement.

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 19 of 25 files reviewed, 6 unresolved discussions (waiting on @brosenberg42, @hhuangMITRE, and @jrobble)

___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 474 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIt0gLS2iUBIqhCskvF:-MIu-os5BgN4h4LysfaK:b-p0ck2l) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L474)):*

Since we are already using JSON in another field, I think it would be better to use JSON here so that consumers don't need to write their own custom parser for this field. You could just put the key-value pairs in a dictionary that is set up like: `{"key1": "value1", "key2": "value2"}` and then use `json.dumps(...)` to convert it to a JSON string.
</blockquote></details>

I also agree it's cleaner than the colon/comma bracket escaped strings. I've switched the output format to a JSON string and also updated the output property name to `KEY_VALUE_PAIRS_JSON` for clarity. 
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 485 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIt89_KCryEuBDEqziD:-MIu067C51SktXisbPBh:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L485)):*

Please add a test for processing `documentResults`. This function has no unit test coverage.
</blockquote></details>

Done.
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 524 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIt-XgSDiRhAOwGTWc1:-MIu09FB9oEIn3OJH-hS:bj3eodx) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L524)):*

This produces invalid JSON. It should be: `fields = json.dumps(doc_result['fields'])`
</blockquote></details>

Fixed, thanks!
___
*[python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py, line 556 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MIsqPSB7_m_6QAfuZsV:-MIu0BOo7CBdJQUM579b:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/d93f6efe791c683d526cf1e927239427fdefb956/python/AzureFormDetection/acs_form_detection_component/acs_form_detection_component.py#L556)):*

Please remove this print statement.
</blockquote></details>

Done.

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: 19 of 25 files reviewed, 6 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[python/AzureFormDetection/README.md, line 39 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/182#-MITyizW8HC1jdLb5Dme:-MIuMgyl13MBGou-ZURI:b-e6lk23) ([raw file](https://github.com/openmpf/openmpf-components/blob/8ae431c60e1a5972ae1febbd0279505970295ebc/python/AzureFormDetection/README.md#L39)):*

Thanks, updated to backticks. Also, corrected the URL description as well. Will update again after making changes to document results.
</blockquote></details>

Documentation updated to reflect current output types.

brosenberg42 (approved):
> Reviewed 6 of 6 files at r7.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/182)* status: all files reviewed, 1 unresolved discussion (waiting on @jrobble)

PR #188, opened 2020-10-07
Updating TesseractOCRTextDetection component to specify custom tessda…

hhuangMITRE:
…ta subdirectory.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/188)

### DIFF
--- a/cpp/TesseractOCRTextDetection/README.md
+++ b/cpp/TesseractOCRTextDetection/README.md
@@ -32,6 +32,10 @@ are stored in `tessdata/script` subdirectory. Users can set a new tessdata direc
 modifying the MODELS_DIR_PATH and TESSDATA_MODELS_SUBDIRECTORY job properties. Once set, the component will look for tessdata files in
 `[MODELS_DIR_PATH]/[TESSDATA_MODELS_SUBDIRECTORY]`.
 
+Please note that existing models will be cached by the component using the model's language and TESSDATA_MODELS_SUBDIRECTORY
+as the identifier. Updating TESSDATA_MODELS_SUBDIRECTORY in a subsequent job run will result in new language models placed
+into the cache.
+
 By default, the component will first check for models in the MODELS_DIR_PATH followed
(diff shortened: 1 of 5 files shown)


brosenberg42:
There is something wrong with this related to caching models. I don't think it checks `TESSDATA_MODELS_SUBDIRECTORY` after the first job runs.  
First, I ran a normal job and it completed as expected. Then, I changed the name of the tessdata directory without changing `TESSDATA_MODELS_SUBDIRECTORY` and ran another job. The job completed successfully even though the tessdata directory was gone. Then, I restarted the component and ran a job. That job failed as expected.

brosenberg42:
This does not fix the problem. I was explaining how I determined that it is using values from the previous job, even though I changed TESSDATA_MODELS_SUBDIRECTORY. The problem is that changing TESSDATA_MODELS_SUBDIRECTORY does nothing for every job other than the first job the component processes.

hhuangMITRE (commented):
Yeah, I also agree with that. I realized that if users wanted to swap to 'tessdata_fast' or 'tessdata_best' then changing the TESSDATA_MODELS_SUBDIRECTORY should force the component to load up a new version of each language file. I've made a change to how we're caching models to take that into account and running some updated unit test to ensure that model reload must occur under these conditions. I'm about to push in the updates now. Thanks!

> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/188#-:-MJDCTDo9zeUznpDlYId:b-hdpcr0)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @brosenberg42 and @jrobble)

jrobble (approved):
> Reviewed 2 of 5 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/188)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

PR #225, opened 2021-01-29
Add OcvYoloDetection component

brosenberg42:
Depends on:
- https://github.com/openmpf/openmpf-cpp-component-sdk/pull/100
- https://github.com/openmpf/openmpf-docker/pull/136

This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)


jrobble:
___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 150 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTI3wiBCpK2zJwAILsH:-MTI3wiC4cZxwmXpGwDD:bwwe7dn) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L150)):*
Possibly because `center` is not exact integer coordinates, From the `getRectSubPix()` doc:

> the values of the pixels at non-integer coordinates are retrieved using bilinear interpolation

It makes me wonder if we're better off ensuring we get an unmodified slice of the frame data that might be a pixel or two off center, width, and/or height, so that we don't have to interpolate.

I ran some tests with the `()` operator with a `Rect2d` roi. Under the hood it's converting the double values to ints with `saturate_cast<int>(double v) { return cvRound(v); }`. Thus, no interpolation.

jrobble (changes requested):
> Reviewed 11 of 57 files at r1, 57 of 59 files at r2, 13 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 51 unresolved discussions (waiting on @brosenberg42)

___
*[cpp/OcvYoloDetection/.dockerignore, line 8 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSix2c33sW3QB4PWwHf:-MSix2c423yfEl8pKWeS:b-53g4lg) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/.dockerignore#L8)):*
> ```ignore
> # 52.227-14, Alt. IV (DEC 2007).                                            #
> #                                                                           #
> # Copyright 2020 The MITRE Corporation. All Rights Reserved.                #
> ```

Please update copyrights.
___
*[cpp/OcvYoloDetection/.gitignore, line 7 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSixQWd-so5H0DBbDkf:-MSixQWe7p9C0YcGakKr:b-ff5a02) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/.gitignore#L7)):*
> ```ignore
> .unotes/
> ObjectDetection-YOLO/
> ocv_face_known_tracks_short.txt
> ```

I don't see `ocv_face_known_tracks_short.txt` or `ObjectDetection-YOLO`. Can those be removed?

Some of the other entries in this section are strange, but I can understand how they might be in Andy's dev env. The last two should probably be under the `# vs-code` section.
___
*[cpp/OcvYoloDetection/Cluster.h, line 83 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj5xSbBJw-QLf7V7CD:-MSj5xSbBJw-QLf7V7CE:b-1zm66e) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Cluster.h#L83)):*
> ```objc
> /** ***************************************************************************
> *  very naive agglomerative clustering (not suitable for lots of items),
> *  move items from a list list of clusters that is returned
> ```

`list` implies `std::list`. We're using `std::vector`. Asking you to change all of the variable names would be too nit-picky. Minimally, I changed some of the comments.
___
*[cpp/OcvYoloDetection/Cluster.h, line 103 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj66WaESy0IpGJG0E1:-MSj66Wb7mJopxGG5ZzV:b-ba5imr) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Cluster.h#L103)):*
> ```objc
>         auto feature = item.getClassFeature();
>         // find 1st cluster item can join
>         // TODO: Should we be using the minimum rather than the first?
> ```

Note to self: Double-check all TODOs before approving.
___
*[cpp/OcvYoloDetection/Cluster.h, line 107 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbYIAv30o1pRIUuLFA:-MTbYIAv30o1pRIUuLFB:b-k3s1c1) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Cluster.h#L107)):*
According to:

> "maximum feature distance an object can be from its cluster's average feature"

Shouldn't this be `distanceFunc(cluster.averageFeature, feature) <= maxDist`?
___
*[cpp/OcvYoloDetection/Config.cpp, line 37 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj91hA92NphXrgwknp:-MSj91hA92NphXrgwknq:buzflak) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Config.cpp#L37)):*
> ```cpp
> using DetectionComponentUtils::GetProperty;
> 
> // Shared static members (might need mutex locks and condition variable if multithreading... )
> ```

This comment does not instill confidence. Does this still need to be addressed?
___
*[cpp/OcvYoloDetection/Config.cpp, line 57 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbfZGWBzk4dDb7DOC9:-MTbfZGWBzk4dDb7DOCA:bhyijhc) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Config.cpp#L57)):*
> ```cpp
>         , maxClassDist(GetProperty(jobProps, "TRACKING_MAX_CLASS_DIST", 0.99))
>         , maxFeatureDist(GetProperty(jobProps, "TRACKING_MAX_FEATURE_DIST", 0.1))
>         , maxCenterDist(GetProperty(jobProps, "TRACKING_MAX_CENTER_DIST", 0.0))
> ```

`TRACKING_MAX_CENTER_DIST` is not in `descriptor.json`. Also, a value of 0.0 indicates that this metric is essentially disabled and has no impact on forming tracks. Is this intentional?
___
*[cpp/OcvYoloDetection/DetectionLocation.h, line 27 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSoI8ai89fTQpn5eFrP:-MSoI8ai89fTQpn5eFrQ:ba5z1ud) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.h#L27)):*
> ```objc
>  ******************************************************************************/
> 
> #ifndef OCVYOLODETECTION_DETECTIONLOCATION_H
> ```

Looking at the existing Darknet component, I see this kind of prefix for these kind of #ifndef lines: `OPENMPF_COMPONENTS_DARKNET*`. 

The .h files in this component are not consistent. Some have `OPENMPF_COMPONENTS_*`, some have `OPENMPF_COMPONENTS_YOLO*`, some have `OPENMPF_COMPONENTS_OCVYOLO*`, and some start with `OCVYOLODETECTION_*`.
___
*[cpp/OcvYoloDetection/DetectionLocation.h, line 37 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSoK7Rv5qBNT6OPLkPH:-MSoK7Rv5qBNT6OPLkPI:b-93j9gw) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.h#L37)):*
> ```objc
> 
> 
> // TODO Should something be done about circular dependency between DetectionLocation and Track?
> ```

Could all of the functions that take a Track object be moved to the Track class and take a Detection object?
___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 105 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTChtFfAGvXy-FvG9N7:-MTChtFg6XIPyIKw6qCp:b-1xc02k) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L105)):*
> ```cpp
> *
> *************************************************************************** */
> cv::Point2d DetectionLocation::phaseCorrelate(Track &track) {
> ```

Very similar to [this](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/phasecorr.cpp#L518). Minimally, we should reference that.
___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 161 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTIIEop5fVLm0W8DFHG:-MTIIEop5fVLm0W8DFHH:bnsekth) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L161)):*
> ```cpp
>     cv::Scalar mean = cv::mean(comp) / 255.0;
>     // combine BGR: d = BGR.BGR
>     // TODO: Why mean.ddot(mean)? That is the vector's length squared.
> ```

Looks like mean squared error to me.
___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 211 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7cyH45p_tVFMQt3CT:-MT7cyH45p_tVFMQt3CU:b1uv641) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L211)):*
> ```cpp
>         if (defaultWindow.empty() || defaultWindow.rows != dftSize_) {
>             //LOG_TRACE("Created default hanning window of size " << defWin.size());
>             cv::createHanningWindow(defaultWindow, cv::Size(dftSize_, dftSize_), CV_32F);
> ```

Instead of `cv::Size(dftSize_, dftSize_)`, can't we just use `size`?
___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 252 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7a_kiD3jjb_11BNGc:-MT7a_kjAOak_8WrWG83:b-bulksm) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L252)):*
> ```cpp
>     cv::meanStdDev(gray, mean, stdDeviation);
>     stdDeviation[0] = std::max(stdDeviation[0], 1 / 255.0);
>     gray.convertTo(gray,
> ```

Note to self: `imshow()` after `convertTo`, once Hanning window applied, and once `dft` is run.
___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 261 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7cCCJE9vy-OhZ1Xva:-MT7cCCL1SkxQzeQwhrg:bsuycjl) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L261)):*
> ```cpp
>     cv::Rect2i grayRoi((gray.size() - feature.size()) / 2, feature.size());
>     // Image  source region
>     grayRoi &= cv::Rect2i(cv::Point2i(0, 0), gray.size());
> ```

Is this to handle the case where `TRACKING_DFT_SIZE` is greater than the frame dimensions?
___
*[cpp/OcvYoloDetection/Dockerfile, line 29 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJG7DO4YxUXiRDRtN-:-MTJG7DO4YxUXiRDRtN0:bhz621q) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Dockerfile#L29)):*
> ```Dockerfile
> #############################################################################
> 
> ARG BUILD_REGISTRY
> ```

Note to self: Perform local Docker build.
___
*[cpp/OcvYoloDetection/Frame.cpp, line 26 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT2hvq-BpMn_xBEkjXq:-MT2hvq-BpMn_xBEkjXr:b-f3ntb1) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Frame.cpp#L26)):*
> ```cpp
>  * limitations under the License.                                             *
>  ******************************************************************************/
> #include "Frame.h"
> ```

Is this file necessary?
___
*[cpp/OcvYoloDetection/KFTracker.cpp, line 81 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT2od8X7PV6__b-5q90:-MT2od8X7PV6__b-5q91:b-h6m34i) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/KFTracker.cpp#L81)):*
> ```cpp
>       // update process noise matrix Q
>       #if PROCESS_NOISE == CONTINUOUS_WHITE
>         _kf.processNoiseCov.at<float>(  i,  i) = _qn.at<float>(b,0) * twentieth_dt5;
> ```

Ask Andy: The content of this file is beyond my current understanding. Is this based on something we can link to or taken from somewhere? How can we verify that this implementation of KF is correct? Unit tests would help here.
___
*[cpp/OcvYoloDetection/KFTracker.cpp, line 178 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT2qtsP1NnlMg9fobeN:-MT2qtsP1NnlMg9fobeO:bs2w44g) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/KFTracker.cpp#L178)):*
> ```cpp
> *************************************************************************** */
> cv::Rect2i KFTracker::_bboxFromState(const cv::Mat1f& state){
>   return cv::Rect2i(static_cast<int>(state.at<float>(0) - state.at<float>(6)/2.0f + 0.5f),
> ```

Ask Andy: Maybe set constants for the indices? Keeping track of what data is in what position is confusing.
___
*[cpp/OcvYoloDetection/KFTracker.cpp, line 372 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT2v_o4BlGlSoFYN5NN:-MT2v_o4BlGlSoFYN5NO:b-jcep6e) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/KFTracker.cpp#L372)):*
> ```cpp
>     _kf.errorCovPost.at<float>(0,0) = rn.at<float>(0);
>     _kf.errorCovPost.at<float>(1,1) = (z0.at<float>(2)/dt) * (z0.at<float>(2)/dt);    // guess max vx as one width/dt
>     _kf.errorCovPost.at<float>(2,2) = 10.0 * _kf.processNoiseCov.at<float>(2,2);      // guess ~3 sigma ?!
> ```

Ask Andy: How does multiplying by 10 relate to ~3 sigma?
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 81 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbg_sWEGMVZjDGTYxo:-MTbg_sWEGMVZjDGTYxp:b-d7eisu) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L81)):*
> ```cpp
>         std::vector<Track> assignedTracks;
>         // intersection over union tracking and assignment
>         Track::assignDetections(trackClusters,
> ```

Implement conditional checks here to skip each of these assign blocks if the respective max distance value is <= 0. Update the descriptions in `descriptor.json` to mention how they can each be disabled.
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 184 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTakLfqEM4Tt8GkEkHl:-MTakLfqEM4Tt8GkEkHm:bh8b5vl) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L184)):*
> ```cpp
>                 }
>             }
>             assignedTracks.insert(assignedTracks.end(),
> ```

Is this just taking unassigned tracks and moving them to assigned tracks? Basically, that would make every unassigned track a new track, right? Leaving a comment explaining as much would help.

Note to self: Check if all track clusters are empty at this time.
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 212 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTakun75qeEVKu8IV1x:-MTakun75qeEVKu8IV1y:beaw2yw) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L212)):*
> "configuration variables are turned into environment variables for late reference"

Is that still true?
___
*[cpp/OcvYoloDetection/ocv_phasecorr.h, line 27 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJDXpeBAkedPX5b4hU:-MTJDXpeBAkedPX5b4hV:b73bt0k) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/ocv_phasecorr.h#L27)):*
> ```objc
>  ******************************************************************************/
> 
> // OpenCV has not exposed these function so we need to duplicate them here :(
> ```

Ask Andy: Provide link to original code.
___
*[cpp/OcvYoloDetection/ocv_phasecorr.cpp, line 38 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJDPnzAf0YR2NnnQN6:-MTJDPnzAf0YR2NnnQN7:b54v9u9) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/ocv_phasecorr.cpp#L38)):*
> ```cpp
> #include <vector>
> 
> namespace cv{
> ```

Ask Andy: Has this code been modified in any way?
___
*[cpp/OcvYoloDetection/README.md, line 8 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj1icB6feHS9dabAGO:-MSj1icCACF53gaIC1Xb:b8kk0vl) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/README.md#L8)):*
> ```Markdown
> ## Trained Models
> 
> The OpenCV Yolo detection component comes with MS COCO trained YoloV4 model:
> ```

If we're going to keep YoloV3 we should mention it in this README.
___
*[cpp/OcvYoloDetection/sample_ocv_yolo_detector.cpp, line 117 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTalxPo4LshQ9kIHkJn:-MTalxPo4LshQ9kIHkJo:b-xg6wve) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/sample_ocv_yolo_detector.cpp#L117)):*
> ```cpp
> 
>     if (argc == 2) {
>         processImage(detection_engine, argc, argv);
> ```

Note to self: Run image test. Run video test.
___
*[cpp/OcvYoloDetection/Track.h, line 207 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbYs9q1BAMzuc_Isdi:-MTbYs9q1BAMzuc_Isdj:bqhcdlx) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.h#L207)):*
According to:

> "maximum feature distance an object can be from its cluster's average feature"

Shouldn't this be `distance <= maxClassDist`?
___
*[cpp/OcvYoloDetection/Track.h, line 244 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT6jCMNA_pZ5CX6YLN4:-MT6jCMO3uM2MMEHqXDt:b-qxx4lh) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.h#L244)):*
> ```objc
> 
>     template<typename TCostFunc>
>     static dlib::matrix<long> getCostMatrix(std::vector<Track> &tracks,
> ```

Could use a function comment.
___
*[cpp/OcvYoloDetection/Track.cpp, line 115 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7L7at3F2127inlOde:-MT7L7au-xBi8uidaHfN:b-3aaeho) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.cpp#L115)):*
> ```cpp
>     if (ocvTracker_.empty()) {   // initialize a new tracker if we don't have one already
>         cv::Rect2i bbox = back().getRect();
>         cv::Rect2i overlap = bbox & cv::Rect2i(0, 0, back().frame.data.cols - 1,
> ```

Leave a comment here saying that we're just checking that some part of the first track detection is within the frame.

Will we have detections that are out of the frame or have a 0 width or height? If so, should we just use the next detection in the track?
___
*[cpp/OcvYoloDetection/Track.cpp, line 119 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7Lao3DIS7Lu2HfAxZ:-MT7Lao47Cs2CLO4Rk2Z:b-6xcz9e) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.cpp#L119)):*
> ```cpp
>         if (overlap.width > 1 && overlap.height > 1) {
>             // could try different trackers here. e.g. cv::TrackerKCF::create();
>             ocvTracker_ = cv::TrackerMOSSE::create();
> ```

Ask Andy: Why was MOSSE selected over the other trackers?
___
*[cpp/OcvYoloDetection/YoloNetwork.cpp, line 54 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJH9121p16j6feiMPV:-MTJH9121p16j6feiMPW:b-j29zfl) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/YoloNetwork.cpp#L54)):*
> ```cpp
>             if (cudaSupportAvailable) {
>                 // A previous job may have been configured to use CUDA, but this job wasn't.
>                 // We call cv::cuda::resetDevice() so that GPU memory used by the previous job
> ```

Note to self: Test switch between GPU and CPU jobs.
___
*[cpp/OcvYoloDetection/YoloNetwork.cpp, line 71 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJHe2g5DoU2yLL7rV4:-MTJHe2g5DoU2yLL7rV5:bhortg0) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/YoloNetwork.cpp#L71)):*
> ```cpp
>         std::string message = "Can not use CUDA because ";
>         if (cudaDeviceCount == 0) {
>             message += "OpenCV was compiled without CUDA support.";
> ```

What if there are simply no GPUs available? See [here](https://android.googlesource.com/platform/external/opencv3/+/master/samples/gpu/cascadeclassifier_nvidia_api.cpp#163). 

Note to self: Test this.
___
*[cpp/OcvYoloDetection/YoloNetwork.cpp, line 245 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJKu427YfFpgTCYjzI:-MTJKu427YfFpgTCYjzJ:b1ws2m8) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/YoloNetwork.cpp#L245)):*
> ```cpp
> 
>         // Convert rectangular image to square image by adding grey bars to the smaller dimensions.
>         // Grey was chosen because that is what the Darknet library does.
> ```

Ask Andy: `DetectionLocation.cpp` does "border replication" when calculating the pixel feature difference, but a gray background is used here. I think that's fine though.
___
*[cpp/OcvYoloDetection/YoloNetwork.cpp, line 246 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJK28k5ouRFPzdA_nP:-MTJK28k5ouRFPzdA_nQ:b-fiu9lv) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/YoloNetwork.cpp#L246)):*
> ```cpp
>         // Convert rectangular image to square image by adding grey bars to the smaller dimensions.
>         // Grey was chosen because that is what the Darknet library does.
>         cv::copyMakeBorder(
> ```

Note to self: `imshow` this.
___
*[cpp/OcvYoloDetection/plugin-files/config/Log4cxxConfig.xml, line 39 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSjAmBS9qFeuR3u13En:-MSjAmBS9qFeuR3u13Eo:b63a165) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/config/Log4cxxConfig.xml#L39)):*
> ```XML
>   </appender>
> 
>   <appender name="Console" class="org.apache.log4j.ConsoleAppender">
> ```

I assume this is left here for debug purposes, or do you think it should be removed?
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 3 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSjAyDJ67f9pSnHXre_:-MSjAyDJ67f9pSnHXrea:b-r5s94v) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L3)):*
> ```JSON
> {
>   "componentName": "OcvYoloDetection",
>   "componentVersion": "1.0",
> ```

Please change the component and middleware version to match the other components when we're ready to move out of beta.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 91 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbasnoEVY4BLq7utFm:-MTbasnoEVY4BLq7utFn:b-aqb47h) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L91)):*
> ```JSON
>           "description": "Threshold class-similarity-distance [0...1] above which detections will not be considered for a track of a particular class.",
>           "type": "FLOAT",
>           "defaultValue": "0.99"
> ```

Refer to other comments about using this value. This description might need to be updated.

Ask Andy: This value seems too high. It seems that this would result in a very small number of clusters.

Note to self: Run test and see how many clusters are formed.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 107 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSmlrJBEmXsMLZuy6WX:-MSmlrJBEmXsMLZuy6WY:bqek9du) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L107)):*
> ```JSON
>         {
>           "name": "TRACKING_EDGE_SNAP_DIST",
>           "description": "Frame border as fraction of frame dimension within which detection bounding boxes are snapped to edge.",
> ```

Ask Andy:

This description could use a sentence that explains why this setting is important, and the implications of setting this too high or low.

It's used a bit in the Kalman filtering code. I can also understand that detections outside the frame region could mess up things like pixel content comparison.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 109 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSjByXG846CdKP5kujU:-MSjByXG846CdKP5kujV:b-xwhwxq) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L109)):*
> ```JSON
>           "description": "Frame border as fraction of frame dimension within which detection bounding boxes are snapped to edge.",
>           "type": "FLOAT",
>           "defaultValue": "0.0075"
> ```

This is 0.005 in Config.cpp. Which should we use?
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 125 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbjA8D-pAPvcLYYruo:-MSjHX0c6NKg5leI2smX:bmhqnil) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L125)):*
> ```JSON
>         {
>           "name": "TRACKING_DFT_USE_HANNING_WINDOW",
>           "description": "Use hanning windowing with DFT (better registration, but slower).",
> ```

Ask Andy: Describe the effect of the Hanning window. Does it essentially remove edge data effects? See [here](https://www.researchgate.net/figure/Weight-functions-to-prepare-the-image-for-FFT-analysis-A-Hanning-window-function_fig1_256478664).
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 143 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSmrZaz68fp_55ju0zq:-MSmrZaz68fp_55ju0zr:bllt5vt) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L143)):*
Note to self:

From Andy (he says it may be wrong):

> If you knock the 100 down to say 50 or 25, it will pay more attention to the measurements than the predictions

Need to verify. Consider updating description.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 176 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSmymBl58ixINFIwV06:-MSmymBl58ixINFIwV07:bp7kg8a) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L176)):*
> ```JSON
>       "properties": [
>         {
>           "name": "CONFIDENCE_THRESHOLD",
> ```

All of the actions use the default `CONFIDENCE_THRESHOLD` of 0.5. Consider removing the property from the actions.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 270 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSmzi783L2cF3vFb6fx:-MSmzi79BQuc62okMhed:b-rtsawa) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L270)):*
> ```JSON
>     },
>     {
>       "name": "OCV YOLO VEHICLE DETECTION ACTION",
> ```

I decided to list all YoloV3 actions, tasks, and pipelines after the YoloV4 ones. Similarly, I listed the tiny variants before the full variants, respectively. Specifically:

```
# actions

OCV TINY YOLO OBJECT DETECTION ACTION
OCV TINY YOLO VEHICLE DETECTION ACTION

OCV YOLO OBJECT DETECTION ACTION
OCV YOLO VEHICLE DETECTION ACTION

OCV TINY YOLO V3 OBJECT DETECTION ACTION
OCV YOLO V3 OBJECT DETECTION ACTION


# tasks

OCV TINY YOLO OBJECT DETECTION TASK
OCV TINY YOLO VEHICLE DETECTION TASK

OCV YOLO OBJECT DETECTION TASK
OCV YOLO VEHICLE DETECTION TASK

OCV TINY YOLO V3 OBJECT DETECTION TASK
OCV YOLO V3 OBJECT DETECTION TASK


# pipelines

OCV TINY YOLO OBJECT DETECTION PIPELINE
OCV TINY YOLO OBJECT DETECTION (WITH MARKUP) PIPELINE

OCV YOLO OBJECT DETECTION PIPELINE
OCV YOLO OBJECT DETECTION (WITH MARKUP) PIPELINE

OCV TINY YOLO V3 OBJECT DETECTION PIPELINE
OCV TINY YOLO V3 OBJECT DETECTION (WITH MARKUP) PIPELINE

OCV YOLO V3 OBJECT DETECTION PIPELINE
OCV YOLO V3 OBJECT DETECTION (WITH MARKUP) PIPELINE
```

Arranging it in this way, it's clear that we're missing the "VEHICLE DETECTION" and "VEHICLE DETECTION (WITH MARKUP)" pipelines for OCV TINY YOLO and OCV YOLO. We have the actions and tasks, but no pipelines that use them.

If we're going to define the actions and tasks here, I think we should also define pipelines that use them here.
___
*[cpp/OcvYoloDetection/plugin-files/models/models.ini, line 19 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj1B1yBTtvnVReE_BD:-MSj1B1yBTtvnVReE_BE:bck6xcn) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/models/models.ini#L19)):*
> ```INI
> 
> 
> [tiny yolo v3]
> ```

Dropping a reminder to make a decision on removing yolov3.
___
*[cpp/OcvYoloDetection/plugin-files/models/person-vehicle-whitelist.txt, line 1 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj-92n4eR4vbuSCPO5:-MSj-92n4eR4vbuSCPO6:b-haxzlk) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/models/person-vehicle-whitelist.txt#L1)):*
> ```vimhelp
> person
> ```

This file is not used. Should we remove it?
___
*[cpp/OcvYoloDetection/test/CMakeLists.txt, line 48 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSn213Q-Ok1QfGsIo3f:-MSn213Q-Ok1QfGsIo3g:b-f5b3sw) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/test/CMakeLists.txt#L48)):*
> ```CMake
> 
>     # allow for 'make check'
>     add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --verbose)
> ```

What's the purpose of this "make check" line?
___
*[cpp/OcvYoloDetection/test/test_ocv_yolo_detection.cpp, line 124 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTaqmg23lIQSM6_YzrD:-MTaqmg3Dh_MX4vEvMPu:bns5avx) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/test/test_ocv_yolo_detection.cpp#L124)):*
> ```cpp
>     double dist = sqrt(diff.dot(diff));
>     GOUT("phase correlation found offset:" << ph_offset << " at a distance of " << dist << " pixels");
>     EXPECT_LE(dist,2.0);
> ```

2.0 is a bit high. The calculated distance for `dog.jpg` is actually  0.771769.

In general, I like testing phase correlation in solution and think we should keep this test.

Ask Andy: It would be good to test the Kalman filter code in isolation too.
___
*[cpp/OcvYoloDetection/plugin-files/models/NOTICE, line 11 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSizTfz8YuwHmdJUepr:-MSizTfz8YuwHmdJUeps:b-s47wft) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/models/NOTICE#L11)):*
> ```
> https://github.com/AlexeyAB/darknet/blob/master/data/coco.names
> 
> ?! Creative Commons Zero v1.0 Universal License http://creativecommons.org/publicdomain/zero/1.0/ ?!
> ```

Ask Andy:

The top level `LICENSE` lists the yolov4 weights and cfg and states:

```
 the resulting product inherits the license and usage restrictions of that model,
 which are not explicitly stated by the original author. Please contact the author
 before using this component for commercial purposes.
```

But in this file we list a `Creative Commons Zero` license.
___
*[cpp/OcvYoloDetection/plugin-files/models/yolov3-confusion-matrix.json, line 1 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj-Jbt-9RQdy6fEWM2:-MSj-Jbt-9RQdy6fEWM3:b-aozunk) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/models/yolov3-confusion-matrix.json#L1)):*
> ```JSON
> {
> ```

Just dropping a comment here to make sure we reach consensus on how we want to move forward with the confusion matrix concept in general after meeting with Andy.
___
*[cpp/OcvYoloDetection/plugin-files/models/yolov3-spp.cfg, line 1 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj-mMkEwPKi6Wlc1Ku:-MSj-mMkEwPKi6Wlc1Kv:b9kjxtm) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/models/yolov3-spp.cfg#L1)):*
> ```INI
> [net]
> ```

This file does not seem to be used. Can we remove it?

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: 23 of 50 files reviewed, 51 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[cpp/OcvYoloDetection/Cluster.h, line 83 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj5xSbBJw-QLf7V7CD:-MTf0aEFDkYBtkIaezW8:b2qvsf2) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Cluster.h#L83)):*

`list` implies `std::list`. We're using `std::vector`. Asking you to change all of the variable names would be too nit-picky. Minimally, I changed some of the comments.
</blockquote></details>

Andy initially used `std::list`. I switched to `std::vector` because the use of `std::list` was causing additional code to be added to deal with the fact that you can't look up items by index.
___
*[cpp/OcvYoloDetection/Cluster.h, line 107 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbYIAv30o1pRIUuLFA:-MTfdHvhDX2szLzBu8Xz:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Cluster.h#L107)):*

According to:

> "maximum feature distance an object can be from its cluster's average feature"

Shouldn't this be `distanceFunc(cluster.averageFeature, feature) <= maxDist`?
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/Config.cpp, line 37 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj91hA92NphXrgwknp:-MTkNl__826j-PlvUXds:b32w12) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Config.cpp#L37)):*

This comment does not instill confidence. Does this still need to be addressed?
</blockquote></details>

No
___
*[cpp/OcvYoloDetection/Config.cpp, line 57 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbfZGWBzk4dDb7DOC9:-MTfdpBa-S3pusmRbyAS:b-h83svt) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Config.cpp#L57)):*

`TRACKING_MAX_CENTER_DIST` is not in `descriptor.json`. Also, a value of 0.0 indicates that this metric is essentially disabled and has no impact on forming tracks. Is this intentional?
</blockquote></details>

That was taken from https://github.com/openmpf/openmpf-components/blob/555b6591e6d61df9afdf45e5be6d5b650d86bb4b/cpp/OcvYoloDetection/plugin-files/config/mpfOcvYoloDetection.ini#L63
___
*[cpp/OcvYoloDetection/DetectionLocation.h, line 27 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSoI8ai89fTQpn5eFrP:-MTf1DI4Ep0PJik25pRD:b-tfynmx) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.h#L27)):*

Looking at the existing Darknet component, I see this kind of prefix for these kind of #ifndef lines: `OPENMPF_COMPONENTS_DARKNET*`. 

The .h files in this component are not consistent. Some have `OPENMPF_COMPONENTS_*`, some have `OPENMPF_COMPONENTS_YOLO*`, some have `OPENMPF_COMPONENTS_OCVYOLO*`, and some start with `OCVYOLODETECTION_*`.
</blockquote></details>

It doesn't matter. It just needs to be a unique string. You can change them if you want.
___
*[cpp/OcvYoloDetection/Frame.cpp, line 26 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT2hvq-BpMn_xBEkjXq:-MTfl0YQAm7YhVgxCsSU:b32w12) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Frame.cpp#L26)):*

Is this file necessary?
</blockquote></details>

No
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 212 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTakun75qeEVKu8IV1x:-MTfrOVTCCYKidchMqWP:b32w12) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L212)):*

> "configuration variables are turned into environment variables for late reference"

Is that still true?
</blockquote></details>

No
___
*[cpp/OcvYoloDetection/Track.h, line 207 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbYs9q1BAMzuc_Isdi:-MTflcFU2KL-GueucZvv:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.h#L207)):*

According to:

> "maximum feature distance an object can be from its cluster's average feature"

Shouldn't this be `distance <= maxClassDist`?
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/YoloNetwork.cpp, line 71 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJHe2g5DoU2yLL7rV4:-MTkQ8VKEnS652PGsMd-:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/YoloNetwork.cpp#L71)):*

What if there are simply no GPUs available? See [here](https://android.googlesource.com/platform/external/opencv3/+/master/samples/gpu/cascadeclassifier_nvidia_api.cpp#163). 

Note to self: Test this.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/YoloNetwork.cpp, line 245 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTJKu427YfFpgTCYjzI:-MTf3fl75r_FZhw6JDUv:b8zg81p) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/YoloNetwork.cpp#L245)):*

Ask Andy: `DetectionLocation.cpp` does "border replication" when calculating the pixel feature difference, but a gray background is used here. I think that's fine though.
</blockquote></details>

Originally, this used border replication. I switched it to do what the darknet library does.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 3 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSjAyDJ67f9pSnHXre_:-MTkQsXV5RTbowIKg3Sw:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L3)):*

Please change the component and middleware version to match the other components when we're ready to move out of beta.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 109 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSjByXG846CdKP5kujU:-MTfo7A0COrEkuFTmRNV:b-r3z1kj) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L109)):*

This is 0.005 in Config.cpp. Which should we use?
</blockquote></details>

The value in Config.cpp was taken from https://github.com/openmpf/openmpf-components/blob/555b6591e6d61df9afdf45e5be6d5b650d86bb4b/cpp/OcvYoloDetection/plugin-files/config/mpfOcvYoloDetection.ini#L66
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 176 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSmymBl58ixINFIwV06:-MTfpVPBE-9K7pPfFCNz:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L176)):*

All of the actions use the default `CONFIDENCE_THRESHOLD` of 0.5. Consider removing the property from the actions.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/plugin-files/models/person-vehicle-whitelist.txt, line 1 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj-92n4eR4vbuSCPO5:-MTfpp-z0xrJqh0eE9M6:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/models/person-vehicle-whitelist.txt#L1)):*

This file is not used. Should we remove it?
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/plugin-files/models/yolov3-spp.cfg, line 1 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSj-mMkEwPKi6Wlc1Ku:-MTfbf7AC2O46cOIiWVK:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/models/yolov3-spp.cfg#L1)):*

This file does not seem to be used. Can we remove it?
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 2 of 59 files at r2, 30 of 30 files at r4.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 26 unresolved discussions (waiting on @brosenberg42)

___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 161 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTIIEop5fVLm0W8DFHG:-MTnV9uUD5urfILm98fh:b22x8zk) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L161)):*

Looks like mean squared error to me.
</blockquote></details>

As discussed, use MAE.
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 81 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbg_sWEGMVZjDGTYxo:-MTnjXPw01kIuIDO9Ql2:b-5sulvh) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L81)):*
Only this remains:

> Implement conditional checks here to skip each of these assign blocks if the respective max distance value is <= 0.
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 184 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTakLfqEM4Tt8GkEkHl:-MTnd8jx24Xe4UdfdosE:b-laoi24) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L184)):*
Only this remains:

> Note to self: Check if all track clusters are empty at this time.
___
*[cpp/OcvYoloDetection/Track.cpp, line 119 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7Lao3DIS7Lu2HfAxZ:-MTnVRlf-CQ5MpbYPmcV:bser0hj) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.cpp#L119)):*

Ask Andy: Why was MOSSE selected over the other trackers?
</blockquote></details>

Because it's fast and accurate.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 91 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbasnoEVY4BLq7utFm:-MTnWeYH7QO7jzvJvrtC:b-vj95o5) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L91)):*
Only this remains:

> Note to self: Run test and see how many clusters are formed.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 109 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSjByXG846CdKP5kujU:-MTnXeu72wvykpC_6KO9:b-iz42nn) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L109)):*

The value in Config.cpp was taken from https://github.com/openmpf/openmpf-components/blob/555b6591e6d61df9afdf45e5be6d5b650d86bb4b/cpp/OcvYoloDetection/plugin-files/config/mpfOcvYoloDetection.ini#L66
</blockquote></details>

We should use the same value in `descriptor.json` and `Config.cpp`. I would assume that Andy did not use the WFM for most of his development, so we should use 0.005 in `descriptor.json`too.

Different values could result in different results when running unit tests and jobs through the WFM.

jrobble (commented):
> Reviewed 15 of 15 files at r5.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 26 unresolved discussions (waiting on @brosenberg42)

jrobble:
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 82 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTo0GMv7xk-oRy-1J7v:-MTo0GMv7xk-oRy-1J7w:b9hawl4) ([raw file](https://github.com/openmpf/openmpf-components/blob/d72bb94621ba185200c3c2b362cac61c6152ed80/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L82)):*
> ```JSON
>         },
>         {
>           "name": "TRACKING_MAX_FRAME_GAP",
> ```

When used by the component, this value should first be divided by the `FRAME_INTERVAL`.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: 45 of 50 files reviewed, 27 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 211 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7cyH45p_tVFMQt3CT:-MTq_esS2p9Bk5JNC-70:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L211)):*

Instead of `cv::Size(dftSize_, dftSize_)`, can't we just use `size`?
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/DetectionLocation.cpp, line 261 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7cCCJE9vy-OhZ1Xva:-MTq_lYgEwhVvGIe08Hi:bq81rta) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/DetectionLocation.cpp#L261)):*

Is this to handle the case where `TRACKING_DFT_SIZE` is greater than the frame dimensions?
</blockquote></details>

Ask Andy
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 81 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTbg_sWEGMVZjDGTYxo:-MTqZSyMD8magIvfa3dC:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L81)):*

Only this remains:

> Implement conditional checks here to skip each of these assign blocks if the respective max distance value is <= 0.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 184 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTakLfqEM4Tt8GkEkHl:-MTfqNAvDamJCKMb4gxi:b-f85paj) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L184)):*

Only this remains:

> Note to self: Check if all track clusters are empty at this time.
</blockquote></details>

I'm not sure if you are referring to `trackCluster.members.empty()` or `trackCluster.members.at(n).locations_.empty()`.
This code does not alter the size of `trackCluster.members`. It is not moving the vector, rather it is moving the individual elements. 
`trackCluster.members.at(n).locations_` is empty because the compiler generated move constructor moves each field. This causes the entire  `trackCluster.members.at(n).locations_`  vector to be moved rather than the individual elements.
___
*[cpp/OcvYoloDetection/Track.cpp, line 115 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7L7at3F2127inlOde:-MTqbi_r7c2wD_OXPSXK:b-dfzyyq) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.cpp#L115)):*

Leave a comment here saying that we're just checking that some part of the first track detection is within the frame.

Will we have detections that are out of the frame or have a 0 width or height? If so, should we just use the next detection in the track?
</blockquote></details>

Detections outside the frame or zero area seems like a bug. I'm not sure what would cause it to be out of the frame.
___
*[cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json, line 109 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MSjByXG846CdKP5kujU:-MTqZmw1BPrQJGbYWZZQ:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/plugin-files/descriptor/descriptor.json#L109)):*

We should use the same value in `descriptor.json` and `Config.cpp`. I would assume that Andy did not use the WFM for most of his development, so we should use 0.005 in `descriptor.json`too.

Different values could result in different results when running unit tests and jobs through the WFM.
</blockquote></details>

Done.

jrobble (commented):
> Reviewed 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 25 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[cpp/OcvYoloDetection/README.md, line 36 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MTsp9JmCWcxm_A98MXr:-MTsp9JnDueGMXb6V3Fz:b-ji5wh5) ([raw file](https://github.com/openmpf/openmpf-components/blob/b92204b8fb2aa445c0a162e4147a74fe75a0e195/cpp/OcvYoloDetection/README.md#L36)):*
> ```Markdown
> * Unassigned detections become new tracks.
> * All active tracks then have their object bounding box positions corrected and predicted forward 
>   in time with a Kalman filter for the next tracking iteration. During the assignment stages, 
> ```

Note to self: Add future research section. Mention potential improvements.

jrobble (commented):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 25 unresolved discussions (waiting on @brosenberg42 and @jrobble)

jrobble (commented):
> Reviewed 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 24 unresolved discussions (waiting on @brosenberg42 and @jrobble)

jrobble (commented):
> Reviewed 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 24 unresolved discussions (waiting on @brosenberg42 and @jrobble)

brosenberg42 (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225#-:-MTw8Zl4AahZgOLlgx89:bg7igkv)* status: 45 of 50 files reviewed, 24 unresolved discussions (waiting on @brosenberg42 and @jrobble)

jrobble (commented):
> Reviewed 5 of 5 files at r10.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/225)* status: all files reviewed, 23 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[cpp/OcvYoloDetection/Track.cpp, line 115 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/225#-MT7L7at3F2127inlOde:-MTw9bZ92Woyr-6nhT1f:b-qcc99a) ([raw file](https://github.com/openmpf/openmpf-components/blob/f4b4733d6cd07e9b0dd135804b00a30639b07978/cpp/OcvYoloDetection/Track.cpp#L115)):*
Only this remains:

> Leave a comment here saying that we're just checking that some part of the first track detection is within the frame.

jrobble (approved):

PR #226, opened 2021-02-24
hotfix/acs-translation-over-char-limit

brosenberg42:
Fixes:
- https://github.com/openmpf/openmpf/issues/1264
- https://github.com/openmpf/openmpf/issues/1260

This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/226)


jrobble (approved):
> Reviewed 25 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/226)* status: all files reviewed, 7 unresolved discussions (waiting on @brosenberg42)

___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 307 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUL0jCv1EK1lNwPUBr-:-MUL0jCv1EK1lNwPUBr0:b-esuscw) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L307)):*
I was going to say:

> This behavior makes sense to me, but is it what the user wants? I'm thinking about the case where we have mixed English and Spanish input and we end up translating everything to Spanish when it's mostly English text. If the JSON output object consumer is only looking at the `TRANSLATION` text, it will be mostly wrong. We might end up feeding mostly garbage to keyword tagging.

> It may be worth providing some indication in the JSON output object that we've chosen an alternative language so that the consumer can ignore that translation if they really just want English. For example, including an additional set of detection properties that start with`ALTERNATIVE TRANSLATION`, leaving the original set as they would be if there were no alternatives. 

But I think I'm overreacting. In the README you provide a Chinese and Spanish example, stating that only one language will be translated. You show that the Chinese is translated but the Spanish is left alone. Is the theory that choosing the alternative language will leave the English alone and only translate the part written in the alternative language? If so, then disregard everything I was going to say since your approach can only improve things, not generate garbage.

Or maybe the detection endpoint falsely returns English as the primary language a lot?

As long as the user is aware that the `TRANSLATION SOURCE` properties will report the alternative non-English language (more precisely, non-`TO_LANGUAGE`), and that's documented in the README, then the current behavior is fine.

However, looking over the chat messages you sent me, you stated that the `TRANSLATION SOURCE` properties will correspond to English when the detected language is English.

...

Please double-check with the user that what you have here is what they want. The big questions in my mind are:

1. Translating using the alternative language costs money. I really like the feature, but if the goal is to get the most bang for our buck then translation should probably only be done using the primary language and the feature should be optional and disabled by default for the user.

2. When the primary language detected is English, does the user always want that reported in `TRANSLATION SOURCE LANGUAGE`? If so, we should probably report the alternative language (and associated confidence) separately.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 332 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUL30jZCB0nmA8FmcZh:-MUL30jZCB0nmA8FmcZi:b-4nc9us) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L332)):*
> ```Python
>         alternative_lang = alternative['language']
>         alternative_lang_score = alternative['score']
>         log.info(f'The detected language, {primary_language}, does not support translation,'
> ```

This makes sense to me, but I think we should report the languages we had to skip (because they don't support translation) in the JSON output object with their associated confidences. This is similar to reporting the missing language models for speech detection.

Similarly, if we keep the behavior to choose an alternative in `_handle_matching_from_and_to_lang()` then we should also consider reporting the skipped languages in the JSON output object.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 415 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUKqfJU-eJzueKp-9-j:-MUKqfJU-eJzueKp-9-k:b-p30jpn) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L415)):*
> ```Python
>             detected_lang_info = response_body[0].get('detectedLanguage')
>             if detected_lang_info:
>                 from_lang = detected_lang_info['language']
> ```

The old component did auto-language detection. Are  you just setting these to pre-inform Azure during the translation call so it doesn't have to perform language detection again?

What if different calls to `_send_break_sentence_request()` result in different detected languages?

In general, I would think that in a feed-forward scenario that each track is a single language and you would not need to worry about this. Maybe only when a text file is provided. It's a corner case.

Thinking about this more, and reading this over again:

```
When the source text is multiple languages, the translation endpoint will only
translate one of the languages ...
```

It seems like what you're doing makes the most sense; that is, translating each chunk using the same "from language", rather than have the first translation call translate all of the Chinese text in the first chunk and the next call translate all of the Spanish text in the next chunk, for example.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 423 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUMKQ-p2EzoXrgAW_ih:-MUMKQ-q5HMN2qj_DkPJ:b-6tgvq0) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L423)):*
> ```Python
>             grouped_sentences.extend(self._process_break_sentence_response(chunk, response_body))
> 
>         log.info('Grouped sentences in to %s chunks.', len(grouped_sentences))
> ```

Grammar: "into"
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 513 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUKyKDu-m8yr0xt37_z:-MUKyKDu-m8yr0xt37a-:bojko6b) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L513)):*
> ```Python
> 
>         punctuation_positions = (text.rfind(ch) for ch in cls.SENTENCE_END_PUNCTUATION)
>         last_punctuation_pos = next((pos for pos in punctuation_positions if pos > 0), -1)
> ```

Is `SENTENCE_END_PUNCTUATION = '.!????'` (note that not all chars display properly here) considered in priority order from left to right? If so, please leave a comment explaining that.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 517 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUL-Ayd1NbfTuaDjrPS:-MUL-Ayd1NbfTuaDjrPT:b-7x7xk8) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L517)):*
> ```Python
>             return last_punctuation_pos + 1
> 
>         last_punctuation_pos = next(
> ```

A comment here explaining that you're looking for the first punctuation, starting from the end of the text, would help.
___
*[python/AzureTranslation/tests/test_acs_translation.py, line 211 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MULE55N5q6gWBi3MHls:-MULE55N5q6gWBi3MHlt:b-69xic9) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/tests/test_acs_translation.py#L211)):*
> ```Python
> 
>         request_body2 = self.get_request_body()
>         self.assertEqual(1, len(request_body1))
> ```

Do you mean to use `len(request_body2)` here?

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/226)* status: 23 of 28 files reviewed, 4 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 307 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUL0jCv1EK1lNwPUBr-:-MUONPV76XOtoCFhziGP:b-ba1omo) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L307)):*

I was going to say:

> This behavior makes sense to me, but is it what the user wants? I'm thinking about the case where we have mixed English and Spanish input and we end up translating everything to Spanish when it's mostly English text. If the JSON output object consumer is only looking at the `TRANSLATION` text, it will be mostly wrong. We might end up feeding mostly garbage to keyword tagging.

> It may be worth providing some indication in the JSON output object that we've chosen an alternative language so that the consumer can ignore that translation if they really just want English. For example, including an additional set of detection properties that start with`ALTERNATIVE TRANSLATION`, leaving the original set as they would be if there were no alternatives. 

But I think I'm overreacting. In the README you provide a Chinese and Spanish example, stating that only one language will be translated. You show that the Chinese is translated but the Spanish is left alone. Is the theory that choosing the alternative language will leave the English alone and only translate the part written in the alternative language? If so, then disregard everything I was going to say since your approach can only improve things, not generate garbage.

Or maybe the detection endpoint falsely returns English as the primary language a lot?

As long as the user is aware that the `TRANSLATION SOURCE` properties will report the alternative non-English language (more precisely, non-`TO_LANGUAGE`), and that's documented in the README, then the current behavior is fine.

However, looking over the chat messages you sent me, you stated that the `TRANSLATION SOURCE` properties will correspond to English when the detected language is English.

...

Please double-check with the user that what you have here is what they want. The big questions in my mind are:

1. Translating using the alternative language costs money. I really like the feature, but if the goal is to get the most bang for our buck then translation should probably only be done using the primary language and the feature should be optional and disabled by default for the user.

2. When the primary language detected is English, does the user always want that reported in `TRANSLATION SOURCE LANGUAGE`? If so, we should probably report the alternative language (and associated confidence) separately.
</blockquote></details>

1. The idea behind using the alternatives is to handle mixed English and foreign text. The English text will be left alone and the foreign text will translated. The user wants this behavior.
2. User wants semicolon delimited list.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 332 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUL30jZCB0nmA8FmcZh:-MUPKsbUD9OCK3Nwe1rG:bpwufho) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L332)):*

This makes sense to me, but I think we should report the languages we had to skip (because they don't support translation) in the JSON output object with their associated confidences. This is similar to reporting the missing language models for speech detection.

Similarly, if we keep the behavior to choose an alternative in `_handle_matching_from_and_to_lang()` then we should also consider reporting the skipped languages in the JSON output object.
</blockquote></details>

I looked in to the set of supported languages. All languages support translation. A subset of those languages support transliteration. I probably should have looked in to it before writing that code.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 415 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUKqfJU-eJzueKp-9-j:-MUOVuzf-uPAcvED6RIp:bvnvro2) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L415)):*
> What if different calls to _send_break_sentence_request() result in different detected languages?

We pass the results from /detect to /breaksentence so /breaksentence does not do language detection.

> In general, I would think that in a feed-forward scenario that each track is a single language and you would not need to worry about this. Maybe only when a text file is provided. It's a corner case.

The problem you are describing only could occur when a single property has multiple languages. The whole track doesn't need to be the same language. Language detection is done independently on each detection.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 513 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUKyKDu-m8yr0xt37_z:-MUOMmIaDkX7mB_m_YhR:bhvv17n) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L513)):*

Is `SENTENCE_END_PUNCTUATION = '.!????'` (note that not all chars display properly here) considered in priority order from left to right? If so, please leave a comment explaining that.
</blockquote></details>

I meant to just find whichever was last in `text`.

jrobble:
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 534 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUPtoC-3R5vdkUu8ocv:-MUPtoC1ErMYvUKT_F0r:b-4go5f9) ([raw file](https://github.com/openmpf/openmpf-components/blob/75212cc352f43837c372c2540d04b16486b00982/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L534)):*
Don't we want to keep this kind of check?

> ```
>             if len(chunk) < max_chars:
>                 result = chunk
> ```

Otherwise, we're generating more chunks than necessary, which results in more calls to `/breaksentence`.

jrobble (commented):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/226)* status: all files reviewed, 3 unresolved discussions (waiting on @brosenberg42)

___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 307 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUL0jCv1EK1lNwPUBr-:-MUPxGxv9HHGzLAiElJB:b-q7rtfj) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L307)):*

1. The idea behind using the alternatives is to handle mixed English and foreign text. The English text will be left alone and the foreign text will translated. The user wants this behavior.
2. User wants semicolon delimited list.
</blockquote></details>

I realized that in most cases the `/detect` call will only return a primary language with no alternatives. I had to try hard to create an example by carefully balancing the amount of English and Chinese text to get an alternative.

What I'm getting at is that in most cases we won't be translating the alternative language since Azure rarely reports one. Thus, the amount of money spent on those calls is relatively minimal. If it was common, we should consider setting a threshold on the confidence value.
___
*[python/AzureTranslation/acs_translation_component/acs_translation_component.py, line 332 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/226#-MUL30jZCB0nmA8FmcZh:-MUPx0wn6Gzu2oR8pdT6:b-wwgil6) ([raw file](https://github.com/openmpf/openmpf-components/blob/f13b2ef005564a0cde70e3aab1b7cc63f34be29b/python/AzureTranslation/acs_translation_component/acs_translation_component.py#L332)):*

I looked in to the set of supported languages. All languages support translation. A subset of those languages support transliteration. I probably should have looked in to it before writing that code.
</blockquote></details>

Ok. A comment in the code explaining that could help avoid confusion in the future.

PR #244, opened 2021-04-19
Feature/nlp text correction

mcrensh:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)


jrobble (changes requested):
I attached `dracula-short.txt.png`. Reviewable doesn't allow direct uploads of `.txt` files, so I had to add the dummy `.png` extension. `NLP TEXT CORRECTION TEXT FILE PIPELINE` took 29 minutes to run on `dracula-short.txt`. Do you know why it takes so long? Is there any way to improve that?



I see that you're using the `lookup_compound()` method to do most of the work. I wonder if it's not designed to take large chunks of text. Instead, I wonder if we should break up the text in sentences based on sentence-ending punctuation (periods, exclamation points, and question marks), run each sentence through independently, and then concatenate the results at the end to get `CORRECTED TEXT`. I know that a sentence is not necessarily the same thing as a phrase, but it gets us much closer.

I'm curious to see what the performance improvement would be when each sentence is run through serially. We can also consider doing sentences in parallel in different threads.

> Reviewed 10 of 10 files at r1, 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244#-:-MYfBqUb25nuDH0uSd-4:b77phq5)* status: all files reviewed, 6 unresolved discussions (waiting on @mcrensh)

___
*[cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json, line 387 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYhBRyq6Y_cAGeuSUoA:-MYhBRyq6Y_cAGeuSUoB:b4c9lyr) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/cpp/TesseractOCRTextDetection/plugin-files/descriptor/descriptor.json#L387)):*
> ```JSON
>     },
>     {
>       "name": "TESSERACT OCR TEXT DETECTION WITH NLP TEXT CORRECTION PIPELINE",
> ```

:thumbsup: 
___
*[python/NlpTextCorrection/README.md, line 6 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYhCIns1VOgeeFZHqVL:-MYhCIns1VOgeeFZHqVM:b-u2n1ws) ([raw file](https://github.com/openmpf/openmpf-components/blob/05ab02c7fa46d93095dc7a9623dc3f82511c0ae8/python/NlpTextCorrection/README.md#L6)):*
> ```Markdown
> component.
> 
> TODO:
> ```

Please complete these TODOs.
___
*[python/NlpTextCorrection/sample_nlp_text_corrector.py, line 38 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYh9ngpA7LNDLv6WiEB:-MYh9ngpA7LNDLv6WiEC:baj9b6v) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/sample_nlp_text_corrector.py#L38)):*
> ```Python
>     _, text = sys.argv
> 
>     job_props = dict(CUSTOM_DICTIONARY='sample_dict.txt')
> ```

I wasn't expecting this to use the custom dictionary. Please make the custom dictionary a separate, optional, command line parameter.

Alternatively, you can scrap this sample script. We're not requiring them anymore since users can make use of the CLI Runner now to test out their component code without the WFM. (Refer to the other comment for how I used it.) It's up to you.

If you choose to scrap it, you can remove the `from .nlp_correction_component ...` line from `__init__.py`.
___
*[python/NlpTextCorrection/setup.py, line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYgtCUF1xV1rNpK5urs:-MYgtCUF1xV1rNpK5urt:b-7r614w) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/setup.py#L36)):*
> ```Python
>         'mpf_component_api>=6.1',
>         'mpf_component_util>=6.1',
>         'numpy>=1.11'
> ```

You're missing a comma here. symspellpy does not get installed via `pip3` without it. I fixed it.

Did you test your Docker image? I don't think it would work without the comma.

---

Also, please add an entry for your component to [this file](https://github.com/openmpf/openmpf-docker/blob/master/docker-compose.components.yml#L45). Note that the components are listed alphabetically. You will need to create a separate PR for that.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 92 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYfDj0o4bAb2ROaiYdb:-MYfDj0o4bAb2ROaiYdc:bbjv4zj) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L92)):*
> ```Python
>                 detection_properties = dict(TEXT=text)
> 
>                 runner = JobRunner(job.job_properties)
> ```

Reuse the `runner` between job runs so that we don't repeatedly incur the cost of loading the dictionaries for every job (which from my count takes over 10 seconds).

Basically, create the `runner` for the first job and save it as a data member until you get the next job. If the dictionary settings are different, then recreate it. Otherwise, use the same one.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 124 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYh738H2vvqm_ahMsN3:-MYh738IBjcKzJm6ephj:bm0y6oq) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L124)):*
> ```Python
>         # load custom dictionary if one is specified in the job properties
>         if custom_dictionary_path != "":
>             full_custom_dict_path = os.path.join(curr_dir, "plugin-files/config/" + custom_dictionary_path)
> ```

Require the user to provide a full path to the custom dictionary file. It should not have to be in `plugin-files/config`. 

Also, check that the file exists. If not, throw an exception with `MPF_COULD_NOT_READ_DATAFILE`.

___

I used this CLI Runner command:

```
docker run --rm -i -v /home/mpf/git/openmpf-projects/openmpf-components/python/NlpTextCorrection/plugin-files/config:/mpfdata openmpf_nlp_text_correction -t generic - -P CUSTOM_DICTIONARY=/mpfdata/sample_dict.txt --pretty < /home/mpf/git/openmpf-projects/openmpf-components/python/NlpTextCorrection/tests/data/sample.txt
```

I expected that to use the custom dictionary, but it didn't. The job completed, meaning it must have silently ignored my attempt to use a custom dictionary. It should have thrown an exception.

For more info on the CLI Runner, refer to the [doc](https://github.com/openmpf/openmpf-docker/blob/master/CLI_RUNNER.md).

In general, the above command mounts a local dir as a volume (`-v`) at `/mpfdata` and uses the `CUSTOM_DICTIONARY` property to tell the component to load `sample_dict.txt` from there.

This kind of command should work if you make the change to use an absolute path.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 135 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYgAdld43LLuCENKqN1:-MYgAdleBhrIimaIOA3Z:bk71doj) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L135)):*
> ```Python
>             original_text, max_edit_distance=2)
> 
>         detection_properties["CORRECTED TEXT"] = suggestions[0].term
> ```

Is it possible to preserve any of the newlines or punctuation?

We may be able to do this ourselves if we break the input text up into separate sentences. See my other comment.

---

Should we enable the option to transfer casing (according to the [doc](https://symspellpy.readthedocs.io/en/latest/api/symspellpy.html#symspellpy.symspellpy.SymSpell.lookup_compound)): `transfer_casing`? I haven't tried using it.

Also, consider enabling the options to keep numbers and acronyms: `ignore_non_words` and `ignore_term_with_digits`. It's worth running some tests to see if those are helpful. Possibly, expose them as algorithm parameters.

Of note, I observed the original text of:

`DATE CHECK AMOUNT\n11/18/11\nTO THE ORDER OF`

changed to the corrected text:

`date check amount of of of to the order of `

As you can see, that completely dropped the date.
___
*[python/NlpTextCorrection/tests/test_nlp_correction.py, line 68 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYhAral4EQ2jbTy8HFZ:-MYhAral4EQ2jbTy8HF_:b-fuelcm) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/tests/test_nlp_correction.py#L68)):*
> ```Python
>         self.assertEqual(1, len(results))
> 
>         expected_text = "this is some sample text explan is misspelled"
> ```

I updated this test to expect "explane". That way we can verify that the custom dictionary actually results in a different word then if the word wasn't corrected at all.
___
*[python/NlpTextCorrection/tests/test_nlp_correction.py, line 69 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYgszb18nCXAtYvH8dc:-MYgszb2Cev2IvbCsr8F:b49q1rj) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/tests/test_nlp_correction.py#L69)):*
> ```Python
> 
>         expected_text = "this is some sample text explan is misspelled"
>         self.assertEqual(expected_text, results[0].detection_properties.get("CORRECTED TEXT"))
> ```

This required adding `main()` to work for me. Otherwise, the test file would silently execute but not actually run the tests. I fixed it.

mcrensh (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: 3 of 14 files reviewed, 6 unresolved discussions (waiting on @jrobble)

___
*[python/NlpTextCorrection/README.md, line 6 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYhCIns1VOgeeFZHqVL:-M_sl9pE0mZuH0CxGElq:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/05ab02c7fa46d93095dc7a9623dc3f82511c0ae8/python/NlpTextCorrection/README.md#L6)):*

Please complete these TODOs.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/sample_nlp_text_corrector.py, line 38 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYh9ngpA7LNDLv6WiEB:-M_sn6s3CVGam3aNnwjb:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/sample_nlp_text_corrector.py#L38)):*

I wasn't expecting this to use the custom dictionary. Please make the custom dictionary a separate, optional, command line parameter.

Alternatively, you can scrap this sample script. We're not requiring them anymore since users can make use of the CLI Runner now to test out their component code without the WFM. (Refer to the other comment for how I used it.) It's up to you.

If you choose to scrap it, you can remove the `from .nlp_correction_component ...` line from `__init__.py`.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/setup.py, line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYgtCUF1xV1rNpK5urs:-M_snsYh-EqXjOKuSehi:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/setup.py#L36)):*

You're missing a comma here. symspellpy does not get installed via `pip3` without it. I fixed it.

Did you test your Docker image? I don't think it would work without the comma.

---

Also, please add an entry for your component to [this file](https://github.com/openmpf/openmpf-docker/blob/master/docker-compose.components.yml#L45). Note that the components are listed alphabetically. You will need to create a separate PR for that.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 92 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYfDj0o4bAb2ROaiYdb:-M_snukCDoWDP0gAZ3Yy:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L92)):*

Reuse the `runner` between job runs so that we don't repeatedly incur the cost of loading the dictionaries for every job (which from my count takes over 10 seconds).

Basically, create the `runner` for the first job and save it as a data member until you get the next job. If the dictionary settings are different, then recreate it. Otherwise, use the same one.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 124 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYh738H2vvqm_ahMsN3:-M_st7J3A5mlQB-oUmMH:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L124)):*

Require the user to provide a full path to the custom dictionary file. It should not have to be in `plugin-files/config`. 

Also, check that the file exists. If not, throw an exception with `MPF_COULD_NOT_READ_DATAFILE`.

___

I used this CLI Runner command:

```
docker run --rm -i -v /home/mpf/git/openmpf-projects/openmpf-components/python/NlpTextCorrection/plugin-files/config:/mpfdata openmpf_nlp_text_correction -t generic - -P CUSTOM_DICTIONARY=/mpfdata/sample_dict.txt --pretty < /home/mpf/git/openmpf-projects/openmpf-components/python/NlpTextCorrection/tests/data/sample.txt
```

I expected that to use the custom dictionary, but it didn't. The job completed, meaning it must have silently ignored my attempt to use a custom dictionary. It should have thrown an exception.

For more info on the CLI Runner, refer to the [doc](https://github.com/openmpf/openmpf-docker/blob/master/CLI_RUNNER.md).

In general, the above command mounts a local dir as a volume (`-v`) at `/mpfdata` and uses the `CUSTOM_DICTIONARY` property to tell the component to load `sample_dict.txt` from there.

This kind of command should work if you make the change to use an absolute path.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 135 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYgAdld43LLuCENKqN1:-M_stBmi5pHJL4TVjn5N:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L135)):*

Is it possible to preserve any of the newlines or punctuation?

We may be able to do this ourselves if we break the input text up into separate sentences. See my other comment.

---

Should we enable the option to transfer casing (according to the [doc](https://symspellpy.readthedocs.io/en/latest/api/symspellpy.html#symspellpy.symspellpy.SymSpell.lookup_compound)): `transfer_casing`? I haven't tried using it.

Also, consider enabling the options to keep numbers and acronyms: `ignore_non_words` and `ignore_term_with_digits`. It's worth running some tests to see if those are helpful. Possibly, expose them as algorithm parameters.

Of note, I observed the original text of:

`DATE CHECK AMOUNT\n11/18/11\nTO THE ORDER OF`

changed to the corrected text:

`date check amount of of of to the order of `

As you can see, that completely dropped the date.
</blockquote></details>

Done.

jrobble:
___
*[python/NlpTextCorrection/setup.py, line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYgtCUF1xV1rNpK5urs:-M_xhaov7-xvVIwV-c7T:b-nukmpz) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/setup.py#L36)):*
> You will need to create a separate PR for that.

Did you create that PR?

mcrensh (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: 3 of 14 files reviewed, 6 unresolved discussions (waiting on @jrobble)

___
*[python/NlpTextCorrection/setup.py, line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MYgtCUF1xV1rNpK5urs:-Ma4cyaQ3QJfKo1cC1mh:by70bj1) ([raw file](https://github.com/openmpf/openmpf-components/blob/b01fbc680e9b33471905d89a2039210321f6ba60/python/NlpTextCorrection/setup.py#L36)):*

> You will need to create a separate PR for that.

Did you create that PR?
</blockquote></details>

The PR is up now

jrobble (changes requested):
> Reviewed 6 of 11 files at r3.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: 9 of 14 files reviewed, 10 unresolved discussions (waiting on @jrobble and @mcrensh)

___
*[python/NlpTextCorrection/Dockerfile, line 40 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5cRWyBpiK2IWFdbKC:-Ma5cRWyBpiK2IWFdbKD:bnxzyci) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/Dockerfile#L40)):*
> ```Dockerfile
>     if [ "${RUN_TESTS,,}" == true ]; then python tests/test_nlp_correction.py; fi
> 
> RUN python -c "import nltk; nltk.download('punkt')"
> ```

This should be `python3`, I think. `pip3` works with `python3`.
___
*[python/NlpTextCorrection/README.md, line 8 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5_QRUCg50AR7Qpg5t:-Ma5_QRUCg50AR7Qpg5u:br3q9i4) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/README.md#L8)):*
> ```Markdown
> # Custom Dictionary
> 
> In order to use a custom dictionary, the full path to a .txt must be included as a job property under the name `CUSTOM_DICTIONARY`. The file should be formatted as a list of word freqeuncy pairs with each word on a separate line. A space should separate each word from its frequency. Here is an example of how the file should be formatted:
> ```

Please explain that the custom dictionary augments (merges with) the default dictionary. It doesn't replace it.

Please explain what happens when the same word is in both dictionaries.
___
*[python/NlpTextCorrection/sample_nlp_text_corrector.py, line 37 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5_0r17bdI8wUhiTop:-Ma5_0r17bdI8wUhiToq:b-gv7qiq) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/sample_nlp_text_corrector.py#L37)):*
> Please make the custom dictionary a separate, optional, command line parameter.

It's a mandatory parameter in your code. To be clear, I meant that your code should accept 1 or 2 parameters. If only the first parameter, the file name, is provided, then the code should use the default dictionary.
___
*[python/NlpTextCorrection/setup.py, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5_xpe5raUihin_LRt:-Ma5_xpe5raUihin_LRu:bpqy9zc) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/setup.py#L34)):*
> ```Python
>     packages=setuptools.find_packages(exclude=('*test*',)),
>     install_requires=(
>         'mpf_component_api>=6.1',
> ```

Please update this line and the one below to 6.2.
___
*[python/NlpTextCorrection/setup.py, line 38 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5dTV95jdTJNUBsJ8O:-Ma5dTV95jdTJNUBsJ8P:btrraa1) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/setup.py#L38)):*
> ```Python
>         'symspellpy>=6.7.0',
>         'setuptools>=8.1',
>         'nltk>=3.5'
> ```

Please mention in the README that the  user must run `python -c "import nltk; nltk.download('punkt')"` after using `pip3` with this `setup.py` file.
___
*[python/NlpTextCorrection/plugin-files/descriptor/descriptor.json, line 3 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5Z-pLDW-7Ffkd6iA6:-Ma5Z-pLDW-7Ffkd6iA7:b-832i7a) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/plugin-files/descriptor/descriptor.json#L3)):*
> ```JSON
> {
>   "componentName": "NlpTextCorrection",
>   "componentVersion": "6.1",
> ```

Please update this line and the next one to 6.2.
___
*[python/NlpTextCorrection/tests/test_nlp_correction.py, line 64 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5ebrAAppgP7wpmWVq:-Ma5ebrAAppgP7wpmWVr:b-zcldrt) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/tests/test_nlp_correction.py#L64)):*
> ```Python
>             data_uri=self._get_test_file("sample.txt"),
>             job_properties=dict(CUSTOM_DICTIONARY='/home/mpf/openmpf-projects/openmpf-components/python'
>                                                   '/NlpTextCorrection/plugin-files/config/sample_dict.txt'),
> ```

I ran this test and got:

```
ERROR:NlpCorrectionComponent:[test-file] Failed to complete job due incorrect file path for the custom dictionary:
NoneType: None
```

Please include the `sample_dict.txt` file.

mcrensh (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: 5 of 15 files reviewed, 10 unresolved discussions (waiting on @jrobble)

___
*[python/NlpTextCorrection/Dockerfile, line 40 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5cRWyBpiK2IWFdbKC:-MaAneVf4odFEhSoEBL0:b-u7i53n) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/Dockerfile#L40)):*

This should be `python3`, I think. `pip3` works with `python3`.
</blockquote></details>

I tried using python3 instead and when I ran the component NLTK was unable to find the punkt data. However, it does run when I use python.
___
*[python/NlpTextCorrection/README.md, line 8 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5_QRUCg50AR7Qpg5t:-MaCM-sT-4UhyY38WWqf:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/README.md#L8)):*

Please explain that the custom dictionary augments (merges with) the default dictionary. It doesn't replace it.

Please explain what happens when the same word is in both dictionaries.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/sample_nlp_text_corrector.py, line 37 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5_0r17bdI8wUhiTop:-MaCM-6S5dEGv2F6K0u4:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/sample_nlp_text_corrector.py#L37)):*

> Please make the custom dictionary a separate, optional, command line parameter.

It's a mandatory parameter in your code. To be clear, I meant that your code should accept 1 or 2 parameters. If only the first parameter, the file name, is provided, then the code should use the default dictionary.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/setup.py, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5_xpe5raUihin_LRt:-MaCLuzj0uQ69xDYPmha:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/setup.py#L34)):*

Please update this line and the one below to 6.2.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/setup.py, line 38 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5dTV95jdTJNUBsJ8O:-MaCLuC5AO1smpDkvGYT:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/setup.py#L38)):*

Please mention in the README that the  user must run `python -c "import nltk; nltk.download('punkt')"` after using `pip3` with this `setup.py` file.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/plugin-files/descriptor/descriptor.json, line 3 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5Z-pLDW-7Ffkd6iA6:-MaCNHD7EJW7Apz6RcAx:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/plugin-files/descriptor/descriptor.json#L3)):*

Please update this line and the next one to 6.2.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/tests/test_nlp_correction.py, line 64 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5ebrAAppgP7wpmWVq:-MaCO4bn9_K8hx2_b5L3:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/tests/test_nlp_correction.py#L64)):*

I ran this test and got:

```
ERROR:NlpCorrectionComponent:[test-file] Failed to complete job due incorrect file path for the custom dictionary:
NoneType: None
```

Please include the `sample_dict.txt` file.
</blockquote></details>

Done.

jrobble:
___
*[python/NlpTextCorrection/Dockerfile, line 40 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Ma5cRWyBpiK2IWFdbKC:-MaWT9TV71gc8hf8u3Bi:b-bj5kjz) ([raw file](https://github.com/openmpf/openmpf-components/blob/0328cb2ef4a655b40390eb327bcc6e47a10bd494/python/NlpTextCorrection/Dockerfile#L40)):*
<details><summary><i>Previously, mcrensh wrote…</i></summary><blockquote>

I tried using python3 instead and when I ran the component NLTK was unable to find the punkt data. However, it does run when I use python.
</blockquote></details>

Never mind. I see that all of the other Python components use `python` to run tests, not `python3`. I didn't pay much attention until now.

jrobble (changes requested):
> Reviewed 2 of 11 files at r3, 7 of 9 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: all files reviewed, 12 unresolved discussions (waiting on @mcrensh)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckSgHQByGFFkK1bExA:-MckSgHQByGFFkK1bExB:b2diku2) (no related file):*
If you run a TESSERACT to NLP pipeline on [this file](https://github.com/openmpf/openmpf-components/blob/master/cpp/TesseractOCRTextDetection/test/data/eng-bul-small.png) the component will generate the following exception:

```
  File "hunspell/hunspell.pyx", line 340, in hunspell.hunspell.HunspellWrap.spell
  File "hunspell/hunspell.pyx", line 80, in hunspell.hunspell.copy_to_c_string
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-4: ordinal not in range(256)
```

That will result in a `UNRECOGNIZED_DETECTION_ERROR`. We should gracefully handle the case where we can't deal with the script / language.

--- 

We may eventually want to consider using [other dictionaries](https://wiki.documentfoundation.org/Development/Dictionaries), and even potentially loading dictionaries from multiple languages at once, but I have no idea if that would mess everything up. For now, let's just focus on English.

---
___
*[python/NlpTextCorrection/README.md, line 3 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj0zp00TpR3Mw3yIJm:-Mcj0zp00TpR3Mw3yIJn:b9mcd63) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/README.md#L3)):*
> ```Markdown
> # Overview
>  
> This repository contains source code and model data for the Natural Language Processing (NLP) text correction component. This component utilizes the symspellpy library, which is a Python port of SymSpell, to perform post-processing correction of OCR text. SymSpell uses Damerau-Levenshtein distance and Baysian probability to correct misspelled text.
> ```

This still mentions symspell.
___
*[python/NlpTextCorrection/README.md, line 10 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj-_H51KYiMley5Oyv:-Mcj-_H51KYiMley5Oyw:bgd8ov0) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/README.md#L10)):*
> the frequency of that word is the frequency in the default dictionary plus the frequency in the custom dictionary.

That's interesting. I imagined that the custom dictionary entries just replace the default ones. It seems that in order to get the frequency numbers right, you would have to generate the custom dictionary with the default dictionary already loaded into memory. Also, it begs the question, is it possible to reduce the frequency of a word already in the default dictionary by having a negative value in the custom dictionary, or is the custom dictionary really just for new words?
___
*[python/NlpTextCorrection/README.md, line 14 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckPLXiDrmEZIt_yK7K:-MckPLXiDrmEZIt_yK7L:b-g4vd73) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/README.md#L14)):*
> ```Markdown
> In order to use a custom dictionary, the full path to a .txt must be included as a job property under the name `CUSTOM_DICTIONARY`. The file should be formatted as a list of word freqeuncy pairs with each word on a separate line. A space should separate each word from its frequency. Here is an example of how the file should be formatted:
> 
> ```the 352
> ```

Is this still accurate? Is doesn't match for format of `sample_dict.dic`:

```
1
explane
```

Please explain this format and how to represent multiple wordss.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 137 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj1n-q9M-e1cuY0JsM:-Mcj1n-q9M-e1cuY0JsN:bl8ad3z) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L137)):*
> ```Python
>             else:
>                 log.exception(f'[{job_name}] '
>                               f'Failed to complete job due incorrect file path for the custom dictionary:')
> ```

Please add the bad file path to this log message in quotes.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 145 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj2HN9EZcvPfdGJkcH:-Mcj2HN9EZcvPfdGJkcI:b1tcal9) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L145)):*
> ```Python
>     # Detection_properties is modified in place.
>     def get_suggestions(self, original_text, detection_properties):
>         log.info(f'Attempting to correct text {original_text}')
> ```

Please make this a debug line or there will be a lot of output in the log file, and use quotes:

```
log.debug(f'Attempting to correct text: "{original_text}"')
```

Also, might has well add a debug line for the corrected text too.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 158 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mck9IUi1uw2FWIPx4Au:-Mck9IUi1uw2FWIPx4Av:bmslsdu) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L158)):*
> ```Python
>         tokenized_sentences = [x.rstrip("\".?!") for x in tokenized_sentences]
> 
>         for sentence, punctuation in zip(tokenized_sentences, punctuation):
> ```

Please use unique variable names. Here `punctuation` is used for both the list and an element from the list.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 164 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj3N_67hTv5Ou1iHk6:-Mcj3N_67hTv5Ou1iHk7:b-171f3i) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L164)):*
> ```Python
> 
>             for word in words:
>                 # if word == "explan":
> ```

Remove debug.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 204 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj7ljA7kOv4l9e0Rfm:-Mcj7ljA7kOv4l9e0Rfn:b1hcvqx) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L204)):*
> ```Python
>                 newline_counter += 2
>             else:
>                 newline_count = len(block) - len(block.rstrip("\n"))
> ```

I think this can be simplified as:

```
newline_count = block.count('\n')
```
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 174 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckFkkO7Nvb4ebhEjiP:-MckFkkO7Nvb4ebhEjiQ:b-yltev6) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L174)):*
> ```Python
>                     temp = self._hunspell.suggest(word)
>                     if len(temp) > 0:
>                         suggestions += self._hunspell.suggest(word)[0]
> ```

Reuse `temp` here.

---

Let's expose a property to optionally enable providing all of the suggestions in the output. For example, I provided this input:

```
This sentance ends in a comman ,
And this one two periods ..
'' Begin with quotes here

** stars **stars ..periods commas,, periods ..

infinite: : a thing,
```

And generated this output:

```
This [sentence, sentience, entrance, stance, senescent] ends in a [conman, comma, command, commas, common, cowman, com man, com-man, comm an, comm-an, comma n, coma] [e, s, i, a, n, r, t, o, l, c, d, u, g, m, p] And this one two periods .. [e's] Begin with quotes here 

** stars [stars] [periods] [commando] periods.

[infinite, infinites, infinitive, indefinite, finite] [e, s, i, a, n, r, t, o, l, c, d, u, g, m, p] a [thing, things, thinning, thinking]
```

I did that by doing `suggestions += '[' + ', '.join(temp) + ']'` where you currently have `suggestions += self._hunspell.suggest(word)[0]`.

That way, a user can increase their odds of finding keywords. Basically, they can error on the side of caution at the risk of false positives. This is also useful for debugging custom dictionaries.

---

Note that in the above generated output the comma at the end of the first line resulted in a suggested "e", and the two periods at the end of the second line resulted in a suggested "e's". Suggestions are also made for the floating ":" in the middle of the last line.

Additionally, there were suggestions for "**stars" and "..periods" because they had leading punctuation. Suggestions are also made for "commas,," due to the trailing punctuation.

It seems that the suggestions are trying to match the number of characters. For example "infinite::::" has a suggestion of "infiniteness".

I'm thinking we should strip leading and trailing punctuation before generating a suggestion (in many cases the word should be found in the dictionary). Also, ignore floating punctuation since a random "e" in the corrected output is wonky.

Add unit tests for these cases.

I'm open to suggestions if you think otherwise.
___
*[python/NlpTextCorrection/plugin-files/descriptor/descriptor.json, line 10 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj16Kh9odxZoFjh4k5:-Mcj16Kh9odxZoFjh4k6:b9mcd63) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/plugin-files/descriptor/descriptor.json#L10)):*
> ```JSON
>   "algorithm": {
>     "name": "NLPTEXTCORRECTION",
>     "description": "Uses the SymSpellPy library to correct text output.",
> ```

This still mentions symspell.
___
*[python/NlpTextCorrection/tests/sample_dict.aff, line 1 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckP7TF5plZZ2e_HrUO:-MckP7TF5plZZ2e_HrUP:bwxbyae) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/tests/sample_dict.aff#L1)):*
> ```
> SET ISO8859-1
> ```

Is this file used for anything?

jrobble (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: all files reviewed, 13 unresolved discussions (waiting on @mcrensh)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckSgHQByGFFkK1bExA:-McoGrawCUceKXwLo4ey:bzpsit) (no related file):*

If you run a TESSERACT to NLP pipeline on [this file](https://github.com/openmpf/openmpf-components/blob/master/cpp/TesseractOCRTextDetection/test/data/eng-bul-small.png) the component will generate the following exception:

```
  File "hunspell/hunspell.pyx", line 340, in hunspell.hunspell.HunspellWrap.spell
  File "hunspell/hunspell.pyx", line 80, in hunspell.hunspell.copy_to_c_string
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-4: ordinal not in range(256)
```

That will result in a `UNRECOGNIZED_DETECTION_ERROR`. We should gracefully handle the case where we can't deal with the script / language.

--- 

We may eventually want to consider using [other dictionaries](https://wiki.documentfoundation.org/Development/Dictionaries), and even potentially loading dictionaries from multiple languages at once, but I have no idea if that would mess everything up. For now, let's just focus on English.
</blockquote></details>

I talked to @brosenberg42 about this. It's best if you wrap this block:

```
                if self._hunspell.spell(word):
                    suggestions += word
                else:
                    temp = self._hunspell.suggest(word)
                    if len(temp) > 0:
                        # suggestions += self._hunspell.suggest(word)[0]
                        suggestions += '[' + ', '.join(temp) + ']'
                    else:
                        suggestions += word
```

with a try/catch so that you can handle the UnicodeEncodeError on a word-by-word basis. If you encounter the exception, just return the original word in the corrected output, then move on to the next word. That way we make an attempt to handle as much text as possible.

It probably makes sense to have one WARN log line stating that you encountered word(s) with unsupported characters. We don't want one log line for each word. That would be too much.

---
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 190 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-McoIFiuBGjyKIQ72lJ5:-McoIFiuBGjyKIQ72lJ6:b-lgtycd) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L190)):*
> ```Python
>         return self._custom_dictionary_path
> 
>     def get_punctuation(self, original_text: str):
> ```

I talked to @brosenberg42 about this. With Symspell it made sense to parse the original text into phrases that could be passed to `lookup_compound()`; however, we're passing individual words into Hunspell. That said, do we have any reason to parse the text into blocks and sentences here? Instead, could we just parse it out word-by-word and keep the leading and trailing punctuation and newline characters that surround each word?

mcrensh (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: 8 of 16 files reviewed, 13 unresolved discussions (waiting on @brosenberg42 and @jrobble)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckSgHQByGFFkK1bExA:-MdugJLh6EflvZwh5jul:b-896fix) (no related file):*

I talked to @brosenberg42 about this. It's best if you wrap this block:

```
                if self._hunspell.spell(word):
                    suggestions += word
                else:
                    temp = self._hunspell.suggest(word)
                    if len(temp) > 0:
                        # suggestions += self._hunspell.suggest(word)[0]
                        suggestions += '[' + ', '.join(temp) + ']'
                    else:
                        suggestions += word
```

with a try/catch so that you can handle the UnicodeEncodeError on a word-by-word basis. If you encounter the exception, just return the original word in the corrected output, then move on to the next word. That way we make an attempt to handle as much text as possible.

It probably makes sense to have one WARN log line stating that you encountered word(s) with unsupported characters. We don't want one log line for each word. That would be too much.
</blockquote></details>

Done.

---
___
*[python/NlpTextCorrection/README.md, line 3 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj0zp00TpR3Mw3yIJm:-MdukTTl77hK6mgHibl1:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/README.md#L3)):*

This still mentions symspell.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/README.md, line 10 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj-_H51KYiMley5Oyv:-MdukOTFC5-JyzlA5l4I:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/README.md#L10)):*

> the frequency of that word is the frequency in the default dictionary plus the frequency in the custom dictionary.

That's interesting. I imagined that the custom dictionary entries just replace the default ones. It seems that in order to get the frequency numbers right, you would have to generate the custom dictionary with the default dictionary already loaded into memory. Also, it begs the question, is it possible to reduce the frequency of a word already in the default dictionary by having a negative value in the custom dictionary, or is the custom dictionary really just for new words?

</blockquote></details>

Done.
___
*[python/NlpTextCorrection/README.md, line 14 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckPLXiDrmEZIt_yK7K:-MdukZoID8qyGceZtOQz:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/README.md#L14)):*

Is this still accurate? Is doesn't match for format of `sample_dict.dic`:

```
1
explane
```

Please explain this format and how to represent multiple wordss.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 137 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj1n-q9M-e1cuY0JsM:-Mduka4s6NIs2bEqKqYc:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L137)):*

Please add the bad file path to this log message in quotes.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 145 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj2HN9EZcvPfdGJkcH:-MdukbfLEcmmsEHLfsiq:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L145)):*

Please make this a debug line or there will be a lot of output in the log file, and use quotes:

```
log.debug(f'Attempting to correct text: "{original_text}"')
```

Also, might has well add a debug line for the corrected text too.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 158 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mck9IUi1uw2FWIPx4Au:-MdukkeG6ba8KSo1C4jr:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L158)):*

Please use unique variable names. Here `punctuation` is used for both the list and an element from the list.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 164 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj3N_67hTv5Ou1iHk6:-MduklqG89eJRMGbjJo_:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L164)):*

Remove debug.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 204 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj7ljA7kOv4l9e0Rfm:-Mdul6GX4GrPKx13VdQ6:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L204)):*

I think this can be simplified as:

```
newline_count = block.count('\n')
```
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 174 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckFkkO7Nvb4ebhEjiP:-MdukptO1Ds6K70cTY1x:b3ab228) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L174)):*

Reuse `temp` here.

---

Let's expose a property to optionally enable providing all of the suggestions in the output. For example, I provided this input:

```
This sentance ends in a comman ,
And this one two periods ..
'' Begin with quotes here

** stars **stars ..periods commas,, periods ..

infinite: : a thing,
```

And generated this output:

```
This [sentence, sentience, entrance, stance, senescent] ends in a [conman, comma, command, commas, common, cowman, com man, com-man, comm an, comm-an, comma n, coma] [e, s, i, a, n, r, t, o, l, c, d, u, g, m, p] And this one two periods .. [e's] Begin with quotes here 

** stars [stars] [periods] [commando] periods.

[infinite, infinites, infinitive, indefinite, finite] [e, s, i, a, n, r, t, o, l, c, d, u, g, m, p] a [thing, things, thinning, thinking]
```

I did that by doing `suggestions += '[' + ', '.join(temp) + ']'` where you currently have `suggestions += self._hunspell.suggest(word)[0]`.

That way, a user can increase their odds of finding keywords. Basically, they can error on the side of caution at the risk of false positives. This is also useful for debugging custom dictionaries.

---

Note that in the above generated output the comma at the end of the first line resulted in a suggested "e", and the two periods at the end of the second line resulted in a suggested "e's". Suggestions are also made for the floating ":" in the middle of the last line.

Additionally, there were suggestions for "**stars" and "..periods" because they had leading punctuation. Suggestions are also made for "commas,," due to the trailing punctuation.

It seems that the suggestions are trying to match the number of characters. For example "infinite::::" has a suggestion of "infiniteness".

I'm thinking we should strip leading and trailing punctuation before generating a suggestion (in many cases the word should be found in the dictionary). Also, ignore floating punctuation since a random "e" in the corrected output is wonky.

Add unit tests for these cases.

I'm open to suggestions if you think otherwise.
</blockquote></details>

With the regex matching leading and trailing punctuation is no longer an issue. Only potential problem would be plural words that end in just an apostrophe.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 190 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-McoIFiuBGjyKIQ72lJ5:-Mdul5RH6pDfnAeV5ILZ:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L190)):*

I talked to @brosenberg42 about this. With Symspell it made sense to parse the original text into phrases that could be passed to `lookup_compound()`; however, we're passing individual words into Hunspell. That said, do we have any reason to parse the text into blocks and sentences here? Instead, could we just parse it out word-by-word and keep the leading and trailing punctuation and newline characters that surround each word?
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/plugin-files/descriptor/descriptor.json, line 10 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj16Kh9odxZoFjh4k5:-Mdua4QE9jCoWsbmAjY7:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/plugin-files/descriptor/descriptor.json#L10)):*

This still mentions symspell.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/tests/sample_dict.aff, line 1 at r5](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MckP7TF5plZZ2e_HrUO:-Mdul2WP-Eb1ClprdKPy:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/717e7dff32443e82baf7db27274c315f66db5939/python/NlpTextCorrection/tests/sample_dict.aff#L1)):*

Is this file used for anything?
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 8 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: all files reviewed, 9 unresolved discussions (waiting on @mcrensh)

___
*[python/NlpTextCorrection/README.md, line 12 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdxcNv545vWJ9J40gB2:-MdxcNv62GvK4rgvXJ0O:b-dykjmx) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/README.md#L12)):*
> ```Markdown
> Using a custom dictionary does not replace the default dictionary. The custom dictionary is merged with the default dictionary. 
> 
> In order to use a custom dictionary, the full path to a .dic must be included as a job property under the name `CUSTOM_DICTIONARY`. The file should be formatted as a list of words, with one word per line. The first line cointains the approximate word count. Here is an example of how the file should be formatted:
> ```

Check spelling: "cointains"
___
*[python/NlpTextCorrection/README.md, line 23 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdxeqxKArMeo2uZT0eI:-MdxeqxKArMeo2uZT0eJ:bz0m7xp) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/README.md#L23)):*
> ```Markdown
> You may also use an affix file (.aff) to specify flags used during spell checking. 
> 
> For more information on dictionary files and affix files for hunspell, see http://manpages.ubuntu.com/manpages/bionic/man5/hunspell.5.html.
> ```

If the content is the same, consider using this link: https://linux.die.net/man/4/hunspell. It's a little strange to be linking to an Ubuntu page when it's not our supported platform.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 145 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj2HN9EZcvPfdGJkcH:-MdxhLWo6KEdMS4BEPFf:b-5emiym) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L145)):*
<details><summary><i>Previously, mcrensh wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Please use quotes around text and add colon, as shown above.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 143 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mdxff3KDddjT27x37Za:-Mdxff3KDddjT27x37Zb:b-hgzlo9) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L143)):*
> ```Python
>             else:
>                 log.exception(f'[{job_name}] '
>                               f'Failed to complete job due incorrect file path for the custom dictionary:'
> ```

Use space after colon and put the bad path in quotes rather than square brackets. That way we're consistent with other components.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 145 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdwHGcU7s7eWP3xNmqi:-MdwHGcVBBFvDizgzrO1:bl99fp5) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L145)):*
> ```Python
>                               f'Failed to complete job due incorrect file path for the custom dictionary:'
>                               f'[{self._custom_dictionary_path}]')
>                 raise mpf.DetectionException(
> ```

Please add the file path in quotes to this exception message.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 154 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdwHY6lDBGlYzgub7Po:-MdwHY6lDBGlYzgub7Pp:b-k4kj6i) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L154)):*
> ```Python
>         log.debug(f'Attempting to correct text {original_text}')
> 
>         suggestions = re.sub(r'[a-zA-Z]+\'?[a-zA-Z]*', self.correct_text, original_text)
> ```

I sent you a separate chat message about how we can go about handling URLs and other specially-formatted words.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 156 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mdxloq32bRzP1A3m3UP:-Mdxloq32bRzP1A3m3UQ:b-bllax4) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L156)):*
> ```Python
>         suggestions = re.sub(r'[a-zA-Z]+\'?[a-zA-Z]*', self.correct_text, original_text)
> 
>         if self._unicode_error:
> ```

You need to reset `unicode_error` for every job. Right now, if you run two jobs back to back, and the first has the error, the second job will generate the warning too, even if there isn't a unicode problem.

I had to force the error by manually setting it in the code. It wasn't generated for Chinese or Cyrillic characters.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 157 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdxlUb73N-4VYlLDysz:-MdxlUb73N-4VYlLDyt-:b-e7blg6) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L157)):*
> ```Python
> 
>         if self._unicode_error:
>             log.warning("Encountered words with unsupported errors")
> ```

Do you mean "unsupported characters"?
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 161 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mdxk8246f9h7bHUJhOE:-Mdxk8246f9h7bHUJhOF:b-9f24ah) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L161)):*
> ```Python
>         detection_properties["CORRECTED TEXT"] = suggestions
>         log.debug("Successfully corrected text")
>         log.debug(f'Corrected text: {suggestions}')
> ```

Please add quotes around text.

mcrensh (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: 13 of 20 files reviewed, 9 unresolved discussions (waiting on @jrobble)

___
*[python/NlpTextCorrection/README.md, line 12 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdxcNv545vWJ9J40gB2:-MexGAcI3qcP-y14cMQL:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/README.md#L12)):*

Check spelling: "cointains"
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/README.md, line 23 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdxeqxKArMeo2uZT0eI:-MexGLOwAWL8U9rjwu9I:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/README.md#L23)):*

If the content is the same, consider using this link: https://linux.die.net/man/4/hunspell. It's a little strange to be linking to an Ubuntu page when it's not our supported platform.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 145 at r4](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mcj2HN9EZcvPfdGJkcH:-MexGf922V-N8U10xtSb:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/e5d4d60e454a1d390cab1747e247ce37257fd72d/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L145)):*

Please use quotes around text and add colon, as shown above.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 143 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mdxff3KDddjT27x37Za:-MexYZ0RBeOx2jtswpI7:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L143)):*

Use space after colon and put the bad path in quotes rather than square brackets. That way we're consistent with other components.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 145 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdwHGcU7s7eWP3xNmqi:-MexYz6aCsRx4p4edzHD:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L145)):*

Please add the file path in quotes to this exception message.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 154 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdwHY6lDBGlYzgub7Po:-MexZ-dF3GzDGoFm3a5f:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L154)):*

I sent you a separate chat message about how we can go about handling URLs and other specially-formatted words.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 156 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mdxloq32bRzP1A3m3UP:-MexX84v20HbuKoqs3Uy:b-11f4in) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L156)):*

You need to reset `unicode_error` for every job. Right now, if you run two jobs back to back, and the first has the error, the second job will generate the warning too, even if there isn't a unicode problem.

I had to force the error by manually setting it in the code. It wasn't generated for Chinese or Cyrillic characters.
</blockquote></details>

Changed it to reset ```unicode_error``` for each job. Also, I get the error with Chinese and Cyrillic characters. 
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 157 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MdxlUb73N-4VYlLDysz:-MexI-Ok0UU4C8eQCxlu:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L157)):*

Do you mean "unsupported characters"?
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 161 at r6](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mdxk8246f9h7bHUJhOE:-MexHu140n-3sQrlabgI:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/8b2088f54a208d263f622fe47fb23c57c4e7cfd8/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L161)):*

Please add quotes around text.
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 7 of 7 files at r7.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: all files reviewed, 3 unresolved discussions (waiting on @mcrensh)

___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 165 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mf-PYXN1fuYHeS0KweS:-Mf-PYXN1fuYHeS0KweT:bnszqx7) ([raw file](https://github.com/openmpf/openmpf-components/blob/b64120904632c04777b618aaf754b624f75c6bcf/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L165)):*
> ```Python
>                     corrected_text = word
>                 else:
>                     temp = self._hunspell.suggest(word)
> ```

Please add the 2-pass behavior we discussed here. As a reminder, if `trailing_punc` starts with a period then concatenate the period to the end of the `word` and see if `_hunspell.spell()` says that's a valid word. If so, use it. If not, get the suggestion without the period.

Please add a test with a custom dictionary with an abbreviation (word that ends in period), so we can make sure the 2-pass behavior is working.
___
*[python/NlpTextCorrection/tests/test_nlp_correction.py, line 111 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mf-LuM18JLOLi9AIJrB:-Mf-LuM18JLOLi9AIJrC:b2517fa) ([raw file](https://github.com/openmpf/openmpf-components/blob/b64120904632c04777b618aaf754b624f75c6bcf/python/NlpTextCorrection/tests/test_nlp_correction.py#L111)):*
I saw `WARNING:NlpCorrectionComponent:Encountered words with unsupported characters` in the log output when I ran this, which is good.

Logging every word with the issue would be too much, so let's just log the first three. Add 

> Encountered words with unsupported characters. The first three (or less) words with the issue are: "one", "two"

The first thing a user reading the original log message is going to want to know is which words have the problem. It's not always going to be obvious if a special character is thrown into a huge chunk of text.
___
*[python/NlpTextCorrection/tests/data/test.txt, line 1 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mf-J1O60C-Pjm8yboP0:-Mf-J1O60C-Pjm8yboP1:beqkf51) ([raw file](https://github.com/openmpf/openmpf-components/blob/b64120904632c04777b618aaf754b624f75c6bcf/python/NlpTextCorrection/tests/data/test.txt#L1)):*
> ```help
> The Project Gutenberg EBook of Dracula, by Bram Stoker
> ```

This is not used anywhere, so please remove it.

mcrensh (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: all files reviewed, 3 unresolved discussions (waiting on @jrobble)

___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 165 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mf-PYXN1fuYHeS0KweS:-MfptiIjBsuMw7DosmZO:b-u9ke2k) ([raw file](https://github.com/openmpf/openmpf-components/blob/b64120904632c04777b618aaf754b624f75c6bcf/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L165)):*

Please add the 2-pass behavior we discussed here. As a reminder, if `trailing_punc` starts with a period then concatenate the period to the end of the `word` and see if `_hunspell.spell()` says that's a valid word. If so, use it. If not, get the suggestion without the period.

Please add a test with a custom dictionary with an abbreviation (word that ends in period), so we can make sure the 2-pass behavior is working.
</blockquote></details>

Changed our approach as discussed
___
*[python/NlpTextCorrection/tests/test_nlp_correction.py, line 111 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mf-LuM18JLOLi9AIJrB:-Mfq1D9h-N7Ox9ZKa3pr:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/b64120904632c04777b618aaf754b624f75c6bcf/python/NlpTextCorrection/tests/test_nlp_correction.py#L111)):*

I saw `WARNING:NlpCorrectionComponent:Encountered words with unsupported characters` in the log output when I ran this, which is good.

Logging every word with the issue would be too much, so let's just log the first three. Add 

> Encountered words with unsupported characters. The first three (or less) words with the issue are: "one", "two"

The first thing a user reading the original log message is going to want to know is which words have the problem. It's not always going to be obvious if a special character is thrown into a huge chunk of text.
</blockquote></details>

Done.
___
*[python/NlpTextCorrection/tests/data/test.txt, line 1 at r7](https://reviewable.io/reviews/openmpf/openmpf-components/244#-Mf-J1O60C-Pjm8yboP0:-Mfq1E5LEF_AhRXNugvg:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/b64120904632c04777b618aaf754b624f75c6bcf/python/NlpTextCorrection/tests/data/test.txt#L1)):*

This is not used anywhere, so please remove it.
</blockquote></details>

Done.

jrobble (commented):
> Reviewed 7 of 7 files at r8, 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: all files reviewed, 1 unresolved discussion (waiting on @mcrensh)

___
*[python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py, line 166 at r8](https://reviewable.io/reviews/openmpf/openmpf-components/244#-MgC5rj40SvBAK_qnSvF:-MgC5rj40SvBAK_qnSvG:b-exyg6n) ([raw file](https://github.com/openmpf/openmpf-components/blob/c9446f32832ab8a77e21cd55090947a8108ffb1a/python/NlpTextCorrection/nlp_correction_component/nlp_correction_component.py#L166)):*
> ```Python
>                     return spacer + leading_punc + word + trailing_punc
> 
>                 word_no_inner_punc = word.translate(str.maketrans("", "", string.punctuation))
> ```

Refactored to only create the table once.

jrobble (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244#-:-MgCLNaD3y6doXelszSI:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @mcrensh)

jrobble (approved):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/244)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mcrensh)

PR #252, opened 2021-05-11
Upgrade trtis-detection component to use the NVIDIA Triton Inference Server

fpertl:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1399
- https://github.com/openmpf/openmpf/issues/1170

**Feature:**

The MPF component `Dockerfile` was updated, but to portion which instantiates an inference server on the fly  and copies over files to run for tests, was never executed.  Development with vscode used separate `.devcontainer/Dockerfile` and `.devcontainer/docker-dompose.yml` not included on github.  These can be provide if that is helpful for the review.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/252)


jrobble:
From a user of the COCO model:

> Hey!!! Heads up, Triton server, object detection, triton will crash if you send it an image that is like 3x > in either dimension than the other, like 300x100 = crash,  100x300 = crash

> I'm not sure what the exact threshold is, but I use 3x as the threshold then pad the short side to be 1/2 of the longest.   crash threshold might be like 3.5x,   I tried 2.9x and it didn't crash.. something that was 3.8x crashed it 

> 442x113 specifically

We should verify this isn't a problem, and if it is, consider padding the input image.

PR #254, opened 2021-05-21
Cnowacki/zero size detection

clnowacki:
**Issues:**
https://github.com/openmpf/openmpf/issues/1308

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/254)

### DIFF
--- a/cpp/OcvYoloDetection/OcvYoloDetection.cpp
+++ b/cpp/OcvYoloDetection/OcvYoloDetection.cpp
@@ -356,14 +356,45 @@ std::vector<MPFVideoTrack> OcvYoloDetection::GetDetections(const MPFVideoJob &jo
             }
         }
 
-        // convert any remaining active tracks to MPFVideoTracks
+        // Convert any remaining active tracks to MPFVideoTracks. Remove any detections
+        // that are below the confidence threshold, and drop empty tracks, if any.
         for (Track &track : inProgressTracks) {
             completedTracks.push_back(Track::toMpfTrack(std::move(track)));
         }
 
-        for (MPFVideoTrack &mpfTrack : completedTracks) {
-            videoCapture.ReverseTransform(mpfTrack);
+        std::vector<std::vector<MPFVideoTrack>::iterator> tracks_to_erase;
+        for (std::vector<MPFVideoTrack>::iterator it = completedTracks.begin();
+             it != completedTracks.end();
+             it++) {
+            MPFVideoTrack &mpfTrack = *it;
+            // Remove detections below the confidence threshold.
+            std::vector<int> detections_to_erase;
+            for (const auto &loc : mpfTrack.frame_locations) {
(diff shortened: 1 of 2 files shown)


jrobble (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/254)* status: all files reviewed, 1 unresolved discussion (waiting on @clnowacki)

___
*[cpp/OcvYoloDetection/Track.cpp, line 188 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/254#-Mb1kYrtEaKa1PAl25Wk:-Mb1kYrtEaKa1PAl25Wl:boe50iy) ([raw file](https://github.com/openmpf/openmpf-components/blob/1a4c8f6972db8c11f775581ed5b3f359f00b0a73/cpp/OcvYoloDetection/Track.cpp#L188)):*
> ```cpp
>                                     back().frame.data.size(), edgeSnap);
>         if ((corrected.width == 0) || (corrected.height == 0)) {
>             kalmanFilterTracker_->setStatePostFromBBox(snapToEdges(back().getRect(), back().getRect(),
> ```

@fpertl, do we need to call `snapToEdges()` again here if we're just reusing what's already at the end of the track? I think that should already have been snapped to edges.

fpertl:
If the snap has already happened in the original assignment, then I don't
think we need to do it again.  In the (hopefully rare) case of zero size
detections, if snap to edges doesn't happen at all it should not really
matter. I think I had left the snap in there from the prior code that
modified the bbox with a correction and would require a new check on the
snap, but as you said the back bbox (unmodified/correcte) should be good to
go as is.

clnowacki (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/254)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @fpertl and @jrobble)

___
*[cpp/OcvYoloDetection/Track.cpp, line 188 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/254#-Mb1kYrtEaKa1PAl25Wk:-MbCvvNU9zXBno4WY861:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/1a4c8f6972db8c11f775581ed5b3f359f00b0a73/cpp/OcvYoloDetection/Track.cpp#L188)):*

@fpertl, do we need to call `snapToEdges()` again here if we're just reusing what's already at the end of the track? I think that should already have been snapped to edges.
</blockquote></details>

Done.

jrobble (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/254)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @clnowacki)

PR #284, opened 2021-09-23
Add Triton support to OcvYoloDetection component

jrobble:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1398
- https://github.com/openmpf/openmpf/issues/1402
- https://github.com/openmpf/openmpf/issues/1465

**Related PRs:**

- https://github.com/openmpf/openmpf-docker/pull/161
- https://github.com/openmpf/openmpf/pull/1474

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/284)


brosenberg42 (changes requested):
> Reviewed 2 of 32 files at r1, 55 of 58 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/284)* status: all files reviewed (commit messages unreviewed), 17 unresolved discussions (waiting on @jrobble)

___
*[cpp/OcvYoloDetection/CMakeLists.txt, line 35 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmpLQr9Pd8ARUBy75E:-MtmpLQr9Pd8ARUBy75F:bk9i6) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/CMakeLists.txt#L35)):*
> ```CMake
> 
> set(CMAKE_BUILD_TYPE RelWithDebInfo)  # build release version with debug symbols
> #set(CMAKE_BUILD_TYPE Release)         # build release version
> ```

The build type should be set to Release.
___
*[cpp/OcvYoloDetection/Config.cpp, line 56 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmqMg-32aFRkqv_-Ua:-MtmqMg-32aFRkqv_-Ub:b-ikxpxp) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/Config.cpp#L56)):*
> ```cpp
> Config::Config(const Properties &jobProps)
>         : confidenceThreshold(std::max(GetProperty(jobProps, "CONFIDENCE_THRESHOLD", 0.5), 0.0)),
>           nmsThresh(GetProperty(jobProps, "DETECTION_NMS_THRESHOLD", 0.3)),
> ```

The common style for constructor initialization lists is to have the leading commas like the code did before this change.
___
*[cpp/OcvYoloDetection/Dockerfile, line 233 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmsYdt8OkW6QknKJf8:-MtmsYdt8OkW6QknKJf9:b-4rp0da) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/Dockerfile#L233)):*
> ```Dockerfile
> COPY --from=build_component $BUILD_DIR/libmpfOcvYoloDetection.so $PLUGINS_DIR/OcvYoloDetection/lib/
> 
> # Copy test files so that they can be used to debug issues later at runtime.
> ```

We really shouldn't be putting things like this in the runtime images. The CLI runner is better for debugging issues at runtime because you can change properties and the media to match a job that is failing.
___
*[cpp/OcvYoloDetection/KFTracker.h, line 39 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmuvAtCaKcZi2jTpht:-MtmuvAuEd_C6qM3F0ah:b-iby17t) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/KFTracker.h#L39)):*
> ```objc
> 
> 
> using namespace std;
> ```

`using namespace ...` should never appear in a header file because it will be applied to everything that `#include`s the header.
___
*[cpp/OcvYoloDetection/KFTracker.h, line 63 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmvkVw9RLwsM9dAo5D:-MtmvkVw9RLwsM9dAo5E:b-osxpu2) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/KFTracker.h#L63)):*
> ```objc
>               const cv::Mat1f &qn);
> 
>     KFTracker(KFTracker &&kft) noexcept: _kf(move(kft._kf)),
> ```

This shouldn't be declared `noexcept` because moving`cv::KalmanFilter` and `cv::Mat1f ` can throw. This can be verified because these produce errors:
```
static_assert(std::is_nothrow_move_constructible<cv::Mat1f>::value, "Is not move constructible");
static_assert(std::is_nothrow_move_constructible<cv::KalmanFilter>::value, "Is not move constructible");
```
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 233 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmyYvt12_C-cG7fNTZ:-MtmyYvt12_C-cG7fNT_:b2cvqbk) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L233)):*
> ```cpp
>     LOG_DEBUG("Initializing OcvYoloDetector");
> 
>     srand(time(NULL));
> ```

Why is this necessary? Using the same seed every time will make the results deterministic.
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 266 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmzSk3EXMnkODxcPio:-MtmzSk4ChxFDyH_nk3R:b1isj8e) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L266)):*
> ```cpp
>     }
> 
>     yoloNetwork_.reset(new YoloNetwork(std::move(modelSettings), config));
> ```

This should not have been changed. As the removed comment states, by making this change, there are temporarily two instances of the yolo network in memory at once. Maybe it isn't obvious initially, but `new YoloNetwork` is executed before `yoloNetwork_.reset` and the old network is in memory until reset is called.  The yolo network is relatively large and it may be in GPU memory which we don't have a lot of.
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 183 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtn_-Wz7VNRkEJuEUBz:-Mtn_-Wz7VNRkEJuEUC-:brzgbxp) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L183)):*
> ```cpp
> 
> 
> bool TritonInferencer::isShmKeyPrefixInUse(const std::string &prefix) {
> ```

You might be able to make this method const.
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 192 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtn74ZmCFzQMri7UHSH:-Mtn74ZmCFzQMri7UHSI:b-vw9efw) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L192)):*
> ```cpp
>                 LOG_WARN("Shared memory prefix \"" << prefix << "\" in use by region on local host: "
>                          << diread->d_name << ". Will try another prefix.");
>                 return true;
> ```

I think you should be calling `closedir(dir);` before returning here.
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 317 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtnaaDZ47ioI9Pl3zA0:-MtnaaDZ47ioI9Pl3zA1:b5fcm4e) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L317)):*
> ```cpp
> 
> void TritonInferencer::waitTillAllClientsReleased() noexcept {
>     if (freeClientIds_.size() != clients_.size()) {
> ```

Accessing `freeClientIds_` here is unsafe. I think you should just remove the conditional here. You don't need to check the sizes between `std::unique_lock<std::mutex> lk(freeClientIdsMtx_);` and `freeClientIdsCv_.wait`. The condition variable will check the lambda before actually waiting. 
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 329 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtnfi_N2XqfOb-x6i1G:-Mtnfi_N2XqfOb-x6i1H:b-jxtao3) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L329)):*
> ```cpp
> 
> void TritonInferencer::rethrowClientException() {
>     if (clientEptr_) {
> ```

I think you need to acquire the mutex here because `clientEptr_` is set by another thread.
___
*[cpp/OcvYoloDetection/triton_server/Dockerfile, line 45 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmjLGk4DFsm0I5Ys9i:-MtmjLGk4DFsm0I5Ys9j:bc99mci) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/Dockerfile#L45)):*
> ```Dockerfile
> RUN --mount=type=tmpfs,target=/tmp \
>     cd /tmp; \
>     git clone https://github.com/openmpf/yolov4-triton-tensorrt.git; \
> ```

In Dockerfiles, I download repos like this: `curl -L https://github.com/openmpf/yolov4-triton-tensorrt/archive/refs/heads/openmpf/dim-608.tar.gz | tar --extract --gzip`.  This is to avoid downloading the repo history and so that we don't have to install git. Git is already installed in this image, but the first point still applies.
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmkkJ67ViplLn7r19q:-MtmkkJ67ViplLn7r19r:b62594g) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L34)):*
> ```Shell
> 
> YOLO_608_PLUGIN_LIB="/plugins/libyolo608layerplugin.so"
> env LD_PRELOAD="$YOLO_608_PLUGIN_LIB" ./gen-engine-608
> ```

The `env` here is not necessary. You can do`MY_VAR=MY_VALUE command` without it. 
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmlAbCEYjE6p-Istsr:-MtmlAbCEYjE6p-Istss:b628umt) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L34)):*
> ```Shell
> 
> YOLO_608_PLUGIN_LIB="/plugins/libyolo608layerplugin.so"
> env LD_PRELOAD="$YOLO_608_PLUGIN_LIB" ./gen-engine-608
> ```

By setting `LD_PRELOAD` here, you are preventing a user from being able to change it at runtime. The default value for `LD_PRELOAD` should be assigned in the Dockerfile.
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 37 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmlQn347fWkrmfoZW9:-MtmlQn347fWkrmfoZWA:b-ibt5vj) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L37)):*
> ```Shell
> 
> # Run the "command" specified in the Docker service entry.
> env /opt/tritonserver/nvidia_entrypoint.sh "$@"
> ```

The leading `env` doesn't do anything, so it should be removed.
___
*[cpp/OcvYoloDetection/yolo_network/BaseYoloNetworkImpl.h, line 48 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtn9GW9CoR6TokgLlpI:-Mtn9GW9CoR6TokgLlpJ:b6rx05j) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/yolo_network/BaseYoloNetworkImpl.h#L48)):*
> ```objc
>     BaseYoloNetworkImpl(ModelSettings modelSettings, const Config &config);
> 
>     ~BaseYoloNetworkImpl();
> ```

If a class has any virtual functions, its destructor should be virtual.
___
*[cpp/OcvYoloDetection/yolo_network/TritonYoloNetworkImpl.cpp, line 52 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtnlXpy0EyExtLBns-C:-MtnlXpy0EyExtLBns-D:b-ctbwel) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/yolo_network/TritonYoloNetworkImpl.cpp#L52)):*
> ```cpp
>     YoloNetworkImpl(ModelSettings model_settings, const Config &config)
>             : BaseYoloNetworkImpl(std::move(model_settings), config),
>               tritonInferencer_(std::move(ConnectTritonInferencer(config))) {}
> ```

In the expression above the return value from `ConnectTritonInferencer` is a temporary value. You shouldn't call `std::move` on temporary values, because it is essentially a no-op in that case. The only possible effect of calling `std::move` on a temporary is preventing rvo.

jrobble (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/284)* status: 41 of 60 files reviewed, 17 unresolved discussions (waiting on @brosenberg42)

___
*[cpp/OcvYoloDetection/CMakeLists.txt, line 35 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmpLQr9Pd8ARUBy75E:-MtoJSix35JzL0C3PuJz:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/CMakeLists.txt#L35)):*

The build type should be set to Release.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/Config.cpp, line 56 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmqMg-32aFRkqv_-Ua:-MtoJuiZ311ex5Ov7Tx3:bj26dvg) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/Config.cpp#L56)):*

The common style for constructor initialization lists is to have the leading commas like the code did before this change.
</blockquote></details>

`Code --> Reformat Code` in CLion changes it to this style by default, but I'll change it back.
___
*[cpp/OcvYoloDetection/Dockerfile, line 233 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmsYdt8OkW6QknKJf8:-MtucdhR0OAtafuFzXmK:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/Dockerfile#L233)):*

We really shouldn't be putting things like this in the runtime images. The CLI runner is better for debugging issues at runtime because you can change properties and the media to match a job that is failing.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/KFTracker.h, line 39 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmuvAtCaKcZi2jTpht:-MtteyO-DAOAj98-r-Us:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/KFTracker.h#L39)):*

`using namespace ...` should never appear in a header file because it will be applied to everything that `#include`s the header.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/KFTracker.h, line 63 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmvkVw9RLwsM9dAo5D:-MtoM0WoEaKmnOMnGakL:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/KFTracker.h#L63)):*

This shouldn't be declared `noexcept` because moving`cv::KalmanFilter` and `cv::Mat1f ` can throw. This can be verified because these produce errors:
```
static_assert(std::is_nothrow_move_constructible<cv::Mat1f>::value, "Is not move constructible");
static_assert(std::is_nothrow_move_constructible<cv::KalmanFilter>::value, "Is not move constructible");
```
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 233 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmyYvt12_C-cG7fNTZ:-MtoMfZGD1OP9ceXKpL_:b-f5ik5d) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L233)):*

Why is this necessary? Using the same seed every time will make the results deterministic.
</blockquote></details>

`rand()` is used here:

```
std::string getRandomShmKeyPrefix() {
    std::stringstream ss;
    ss << "/OcvYoloDetection_" << hostname() << "_" << std::setw(10) << std::setfill('0') << rand();
    return ss.str();
}
```

`time(NULL)` returns the number of seconds since the epoch. That will generally result in a unique seed. Here's what I get on two separate runs:

```
t: 1642630652
t: 1642630699
```
___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 266 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmzSk3EXMnkODxcPio:-MtoQiIA1bV8xid48fZX:bqwjjpa) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L266)):*

This should not have been changed. As the removed comment states, by making this change, there are temporarily two instances of the yolo network in memory at once. Maybe it isn't obvious initially, but `new YoloNetwork` is executed before `yoloNetwork_.reset` and the old network is in memory until reset is called.  The yolo network is relatively large and it may be in GPU memory which we don't have a lot of.
</blockquote></details>

I was not aware. It seemed redundant to call `reset()` twice, but what you said about keeping two instances in memory makes sense.
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 183 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtn_-Wz7VNRkEJuEUBz:-MtoReM7ED2U7jWYloC_:bggs6yu) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L183)):*

You might be able to make this method const.
</blockquote></details>

I can. Done.
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 192 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtn74ZmCFzQMri7UHSH:-MtoRlna9YXxyNkCCgEi:b6zrrkh) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L192)):*

I think you should be calling `closedir(dir);` before returning here.
</blockquote></details>

Good catch.
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 317 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtnaaDZ47ioI9Pl3zA0:-MtoT6KM8R-zMyMLefGr:bb6un1j) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L317)):*

Accessing `freeClientIds_` here is unsafe. I think you should just remove the conditional here. You don't need to check the sizes between `std::unique_lock<std::mutex> lk(freeClientIdsMtx_);` and `freeClientIdsCv_.wait`. The condition variable will check the lambda before actually waiting. 
</blockquote></details>

Done
___
*[cpp/OcvYoloDetection/triton/TritonInferencer.cpp, line 329 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtnfi_N2XqfOb-x6i1G:-MtoTkM17jlgW_EH6Hi4:b-essr4o) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L329)):*

I think you need to acquire the mutex here because `clientEptr_` is set by another thread.
</blockquote></details>

This function is only ever called here:

```
    void Finish() override {
        if (tritonInferencer_) {
            // wait for clients and check for client exception at the end of the job
            tritonInferencer_->waitTillAllClientsReleased();
            frameIdxComplete_ = -1;
            tritonInferencer_->rethrowClientException();
        }
    }
```

I was going to say that because `rethrowClientException` occurs after `waitTillAllClientsReleased` that it should not be possible for another thread to set `clientEptr_`, until I realized that:

```
void TritonInferencer::releaseClientId(int clientId, const std::exception_ptr& newClientEptr) {
    {
        std::lock_guard<std::mutex> lk(freeClientIdsMtx_);
        freeClientIds_.insert(clientId);
        if (newClientEptr && !clientEptr_) {
            clientEptr_ = newClientEptr;
        }
        LOG_TRACE("Freeing client[" << clientId << "]");
    }
    freeClientIdsCv_.notify_all();
}
```

Should be changed to:

```
void TritonInferencer::releaseClientId(int clientId, const std::exception_ptr& newClientEptr) {
    {
        std::lock_guard<std::mutex> lk(freeClientIdsMtx_);
        if (newClientEptr && !clientEptr_) {
            clientEptr_ = newClientEptr;
        }
        freeClientIds_.insert(clientId);
        LOG_TRACE("Freeing client[" << clientId << "]");
    }
    freeClientIdsCv_.notify_all();
}
```

Note how I moved `freeClientIds_.insert(clientId);` down.

I committed that change.
___
*[cpp/OcvYoloDetection/triton_server/Dockerfile, line 45 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmjLGk4DFsm0I5Ys9i:-MtoVZVC4ElUXzrZfm3b:b-3r8jmv) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/Dockerfile#L45)):*

In Dockerfiles, I download repos like this: `curl -L https://github.com/openmpf/yolov4-triton-tensorrt/archive/refs/heads/openmpf/dim-608.tar.gz | tar --extract --gzip`.  This is to avoid downloading the repo history and so that we don't have to install git. Git is already installed in this image, but the first point still applies.
</blockquote></details>

I'll make this change since we're only supporting 608x608 images now. The clone made more sense when I was also building the 416x416 plugin using a different branch of the repo.
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmkkJ67ViplLn7r19q:-MtuZ2EoCd9OtM0W6v61:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L34)):*

The `env` here is not necessary. You can do`MY_VAR=MY_VALUE command` without it. 
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmlAbCEYjE6p-Istsr:-MttjdtcESWkglC2B9cF:bitqa53) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L34)):*

By setting `LD_PRELOAD` here, you are preventing a user from being able to change it at runtime. The default value for `LD_PRELOAD` should be assigned in the Dockerfile.
</blockquote></details>

I removed the `env`. Otherwise, it's currently being set in the `docker-compose.yml` service entry:

```
    environment:
      # Optionally, limit the GPUs exposed to the server.
      # At least one GPU is required to run the YOLO TensorRT engine.
      - NVIDIA_VISIBLE_DEVICES=all
      - LD_PRELOAD=/plugins/libyolo608layerplugin.so
```

One reason it's set there is because `--load-model=yolo-608` is set there. I think it's important for people to know that the plugin is used when the model is used.

In the future we might support other models in the same Triton instance with their own custom plugin libraries. I don't want anyone to provide a different `LD_PRELOAD` value in the service entry that omits the 608 plugin. If they see the env variable already set there, then they should know enough to just append whatever they need to the end of it.

I'll go ahead and set `ENV LD_PRELOAD` in the Dockerfile too, so in case anyone removes `LD_PRELOAD` from the service entry it will still work. 
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 37 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmlQn347fWkrmfoZW9:-MtuZ-oQCESdPz-fqUj6:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L37)):*

The leading `env` doesn't do anything, so it should be removed.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/yolo_network/BaseYoloNetworkImpl.h, line 48 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-Mtn9GW9CoR6TokgLlpI:-Mttf8DF6gX9lzq-EubO:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/yolo_network/BaseYoloNetworkImpl.h#L48)):*

If a class has any virtual functions, its destructor should be virtual.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/yolo_network/TritonYoloNetworkImpl.cpp, line 52 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtnlXpy0EyExtLBns-C:-MttginO7AJurZjkz2K_:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/yolo_network/TritonYoloNetworkImpl.cpp#L52)):*

In the expression above the return value from `ConnectTritonInferencer` is a temporary value. You shouldn't call `std::move` on temporary values, because it is essentially a no-op in that case. The only possible effect of calling `std::move` on a temporary is preventing rvo. 
</blockquote></details>

Done.

brosenberg42 (changes requested):
> Reviewed 19 of 19 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/284)* status: all files reviewed, 1 unresolved discussion (waiting on @jrobble)

___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 233 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmyYvt12_C-cG7fNTZ:-Mtx0pPV8_ikdJ5iNA8Q:b2dpy6c) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L233)):*

`rand()` is used here:

```
std::string getRandomShmKeyPrefix() {
    std::stringstream ss;
    ss << "/OcvYoloDetection_" << hostname() << "_" << std::setw(10) << std::setfill('0') << rand();
    return ss.str();
}
```

`time(NULL)` returns the number of seconds since the epoch. That will generally result in a unique seed. Here's what I get on two separate runs:

```
t: 1642630652
t: 1642630699
```
</blockquote></details>

My concern here is that `srand` is a global change and parts of OpenCV might use `rand()`, so we might be introducing non-deterministic behavior when we don't need to be. The other thing is that if you really want unique seeds, one second is not precise enough. It is relatively likely that during a `docker stack deploy` multiple services will be started within a second. `std::random` lets you create a non-global rng. The example here: http://www.cplusplus.com/reference/random/uniform_int_distribution/uniform_int_distribution/ is pretty good except I would change it to use  `std::chrono::high_resolution_clock` instead of ` std::chrono::system_clock`.
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmlAbCEYjE6p-Istsr:-Mtx3Qa0DTbyW1duT3oh:bus9lsl) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L34)):*
> I don't want anyone to provide a different LD_PRELOAD value in the service entry that omits the 608 plugin. 

If you wanted to do that, you would need something like:
`LD_PRELOAD=$LD_PRELOAD:/plugins/libyolo608layerplugin.so`

jrobble (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/284)* status: 57 of 60 files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

___
*[cpp/OcvYoloDetection/OcvYoloDetection.cpp, line 233 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmyYvt12_C-cG7fNTZ:-MtxmJjm0oMcGG6zOzTq:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/OcvYoloDetection.cpp#L233)):*

My concern here is that `srand` is a global change and parts of OpenCV might use `rand()`, so we might be introducing non-deterministic behavior when we don't need to be. The other thing is that if you really want unique seeds, one second is not precise enough. It is relatively likely that during a `docker stack deploy` multiple services will be started within a second. `std::random` lets you create a non-global rng. The example here: http://www.cplusplus.com/reference/random/uniform_int_distribution/uniform_int_distribution/ is pretty good except I would change it to use  `std::chrono::high_resolution_clock` instead of ` std::chrono::system_clock`.
</blockquote></details>

Done.
___
*[cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh, line 34 at r3](https://reviewable.io/reviews/openmpf/openmpf-components/284#-MtmlAbCEYjE6p-Istsr:-Mty3ivX7UJ9BA7bIECT:b67faki) ([raw file](https://github.com/openmpf/openmpf-components/blob/be361d579ee6dfbed4b5010619ae3e41dc5b8dae/cpp/OcvYoloDetection/triton_server/docker-entrypoint.sh#L34)):*

> I don't want anyone to provide a different LD_PRELOAD value in the service entry that omits the 608 plugin. 

If you wanted to do that, you would need something like:
`LD_PRELOAD=$LD_PRELOAD:/plugins/libyolo608layerplugin.so`
</blockquote></details>

As discussed:

Let's not do the `export`. I'll leave `ENV LD_PRELOAD /plugins/libyolo608layerplugin.so` in the `Dockerfile` since that will let things work if the user for some reason omits `LD_PRELOAD` from the service entry. This approach will enable them to completely override that default value by providing something else in the `LD_PRELOAD` service entry (they will also need to omit `--load-model=yolo-608` or there will be problems, but that's on them).

brosenberg42 (approved):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/284)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jrobble)

brosenberg42 (approved):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/284)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jrobble)

PR #305, opened 2022-09-12
Adding support for OpenNLP language detector and updating TikaTextDetection to output multiple language predictions.

hhuangMITRE:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1592
- https://github.com/openmpf/openmpf/issues/1545

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/305)

### DIFF
--- a/java/TikaTextDetection/LICENSE
+++ b/java/TikaTextDetection/LICENSE
@@ -27,3 +27,9 @@ The "language-detector" project is licensed under the Apache 2.0 License.
 Project Page: https://github.com/optimaize/language-detector
 
 --------------------------------------------------------------------------
+
+The "Apache OpenNLP" project is licensed under the Apache 2.0 License.
+Project Page: https://github.com/apache/opennlp
+
+--------------------------------------------------------------------------
+
--- a/java/TikaTextDetection/README.md
+++ b/java/TikaTextDetection/README.md
@@ -40,7 +40,7 @@ The following format-specific behaviors were observed using Tika 1.28.1 on Ubunt
 - OpenDocument Spreadsheet documents will generate one track per cell, as well as some additional tracks with
   date and time information, "Page /", and "???".
 
-# Language detection parameters
+# Language Detection Parameters
 
 Tika supports the following language detection properties:
 
@@ -60,10 +60,11 @@ Please note that the behavior of `MIN_LANGUAGES` is different depending on the l
 Language results are stored as follows:
 - `TEXT_LANGUAGE` : The primary detected language for a given text. Set to "Unknown" if no language is identified.
 - `TEXT_LANGUAGE_CONFIDENCE` : A confidence setting for the primary language ranging from `NONE` to `HIGH` confidence. See [note here.](https://tika.apache.org/1.21/api/org/apache/tika/language/detect/LanguageConfidence.html)
+- `ISO_LANGUAGE` : The primary detected language for a given text in ISO 639-3 format. Set to "UNKNOWN" if no language id identified (as "UNK" is an ISO code).
 
 Secondary languages and their confidence scores are listed as comma separated strings:
 - `SECONDARY_TEXT_LANGUAGES` : A list of secondary languages (from greatest to least confidence) separated by ", " delimiters.
-    Set to "Unknown" if no language is identified.
+    Set to an empty string if no language is identified.
 - `SECONDARY_TEXT_LANGUAGE_CONFIDENCES` : A confidence list corresponding to the secondary detected languages in order (also separated by commas). 
 
 # Supported Language Detectors:
@@ -103,128 +104,128 @@ the `LANG_DETECTOR` option:
   - Belarusian
   - Catalan
   - Danish
-  - German
+  - Dutch
+  - English
   - Esperanto
   - Estonian
-  - Greek
-  - English
-  - Spanish
   - Finnish
   - French
-  - Persian
   - Galician
+  - German
+  - Greek
   - Hungarian
   - Icelandic
   - Italian
   - Lithuanian
-  - Dutch
   - Norwegian
+  - Persian
   - Polish
   - Portuguese
   - Romanian
   - Russian
   - Slovakian
   - Slovenian
+  - Spanish
   - Swedish
   - Thai
   - Ukrainian
-                 
+
 
 - Optimaize Language Detector:
   - Every language in Tika's language detector, except Esperanto.
   - Afrikaans
   - **Aragonese** (Unique to this detector)
+  - Albanian
   - Arabic
   - Asturian
+  - Basque
+  - Bengali
   - Breton
   - Bulgarian
-  - Bengali
+  - Croatian
   - Czech
-  - Welsh
-  - Basque
-  - Irish
   - Gujarati
+  - Haitian
   - Hebrew
   - Hindi
-  - Croatian
-  - Haitian
   - Indonesian
+  - Irish
   - Japanese
-  - Khmer
   - Kannada
+  - Khmer
   - Korean
   - Latvian
   - Macedonian
-  - Malayalam
-  - Marathi
   - Malay
+  - Malayalam
   - Maltese
+  - Marathi
   - Nepali
   - Occitan
   - Punjabi
+  - Serbian
+  - Simplified Chinese
   - Slovak
   - Slovene
   - Somali
-  - Albanian
-  - Serbian
   - Swahili
+  - Tagalog
   - Tamil
   - Telugu
-  - Tagalog
+  - Traditional Chinese
   - Turkish
   - Urdu
   - Vietnamese
   - Walloon
+  - Welsh
   - Yiddish
-  - Simplified Chinese
-  - Traditional Chinese
                         
 
 - OpenNLP Language Detector:
   - Every language in Tika and Optimaize Language Detectors except Aragonese. 
-  - Bihari languages 
-  - Swiss German
-  - Turkmen
-  - Bashkir
-  - Mongolian
-  - Balinese
-  - Pushto
-  - Faroese
-  - Swati
-  - Min Nan Chinese
-  - Yoruba
-  - Scottish Gaelic
-  - Javanese
-  - Iranian Persian
-  - Esperanto
-  - Western Panjabi
-  - Standard Latvian
-  - Western Frisian
-  - Burmese
-  - Eastern Mari
-  - Paraguayan Guaraní
-  - Slovenian
-  - Cebuano
-  - Mandarin Chinese
-  - Kurdish
-  - Pedi
-  - Azerbaijani
-  - Uighur
-  - Minangkabau
-  - Tajik
-  - Uzbek
-  - Maori
-  - Sindhi
-  - Konkani
   - Armenian
-  - Igbo
   - Assamese
-  - Malay
-  - Low German
-  - Fulah
-  - Xhosa
-  - Standard Estonian
-  - Goan Konkani
-  - Lingala
+  - Azerbaijani
+  - Balinese
+  - Bashkir
+  - Bihari languages
+  - Burmese
+  - Cebuano
   - Dhivehi
+  - Eastern Mari
+  - Esperanto
+  - Faroese
+  - Fulah
+  - Goan Konkani
+  - Igbo
+  - Iranian Persian
+  - Javanese
+  - Konkani
+  - Kurdish
+  - Lingala
+  - Low German
+  - Malay
+  - Mandarin Chinese
+  - Maori
+  - Min Nan Chinese
+  - Minangkabau
+  - Mongolian
+  - Paraguayan Guaraní
+  - Pedi
+  - Pushto
+  - Scottish Gaelic
+  - Sindhi
+  - Slovenian
+  - Standard Estonian
+  - Standard Latvian
+  - Swati
+  - Swiss German
+  - Tajik
+  - Turkmen
+  - Uighur
+  - Uzbek
+  - Western Frisian
+  - Western Panjabi
+  - Xhosa
+  - Yoruba
   - Zulu
--- a/java/TikaTextDetection/README.md
+++ b/java/TikaTextDetection/README.md
@@ -48,8 +48,6 @@ Tika supports the following language detection properties:
 
 For instance, if `MIN_LANGUAGES` is set to 2, the component will always attempt to return the top 2 predicted languages, followed by the next languages if they are also marked as reasonably certain.
 
-If `MAX_REASONABLE_LANGUAGES` is set to -1 and `MIN_LANGUAGES` is set to 2 (default), the component will always attempt to return the top predicted language and a secondary language, even if they are not set to reasonably confident by Tika.
-
 Please note that the behavior of `MIN_LANGUAGES` is different depending on the language detector:
 - The Optimaize language detector will often produce the exact number of `MIN_LANGUAGES` requested by a user.
 - The OpenNLP language detector tends to only produce 1 language unless `MIN_LANGUAGES` is set to 2 or greater. However, low confidence results are still sometimes thrown out even if the user requested additional language predictions. (This indicates that OpenNLP uses a different filtering mechanism which could be investigated further).
@@ -69,9 +67,9 @@ Secondary languages and their confidence scores are listed as comma separated st
 # Supported Language Detectors:
 
 This component supports the following language detectors. Users can select their preferred detector using
-the `LANG_DETECTOR` option:
+the `LANGUAGE_DETECTOR` option:
                                                                                                                                                                           
-- `LANG_DETECTOR = opennlp`: [Apache Tika OpenNLP Language Detector](https://tika.apache.org/2.4.1/api/org/apache/tika/langdetect/opennlp/OpenNLPDetector.html)           
+- `LANGUAGE_DETECTOR = opennlp`: [Apache Tika OpenNLP Language Detector](https://tika.apache.org/2.4.1/api/org/apache/tika/langdetect/opennlp/OpenNLPDetector.html)           
                                                                                                                                                                           
   Apache Tika's latest in-house language detection capability based on OpenNLP's language detector.                                                                       
   Uses Machine Learning (ML) models trained on the following datasets and supports 148 languages in total                                                                   
@@ -82,7 +80,7 @@ the `LANG_DETECTOR` option:
 
   **PLEASE NOTE**, if `opennlp` is selected, ensure that the `MIN_LANGUAGES` property is set to 1 or greater.
 
-- `LANG_DETECTOR = optimaize`: [Optimaize Language Detector](https://github.com/optimaize/language-detector)
+- `LANGUAGE_DETECTOR = optimaize`: [Optimaize Language Detector](https://github.com/optimaize/language-detector)
 
   Third party language detection project that supports 71 languages.
   Predicts target language using N-gram frequency matching between input and language profiles.
--- a/java/TikaTextDetection/README.md
+++ b/java/TikaTextDetection/README.md
@@ -40,6 +40,32 @@ The following format-specific behaviors were observed using Tika 1.28.1 on Ubunt
 - OpenDocument Spreadsheet documents will generate one track per cell, as well as some additional tracks with
   date and time information, "Page /", and "???".
 
+# Language detection parameters
+
+Tika supports the following language detection properties:
+
+- `MAX_REASONABLE_LANGUAGES`: Specifies maximum number of top detected languages. 
+When set to 0 or below, allow any number of language results that are marked as reasonably certain by Tika.
+
+- `MIN_LANGUAGES`: When set to a positive integer, attempt to return specified number of top languages, even if some are not marked as reasonably certain. Non-positive values disable this property to only accept reasonable predictions.
+
+For instance, if `MAX_REASONABLE_LANGUAGES` is set to 5 and `MIN_LANGUAGES` is set to 2, the component will always attempt to return the top 2 predicted languages, followed by the next 3 if they are marked as reasonably certain.
+
+If `MAX_REASONABLE_LANGUAGES` is set to -1 and `MIN_LANGUAGES` is set to 2 (default), the component will always attempt to return the top predicted language and a secondary language, even if they are not set to reasonably confident by Tika.
+
+Please note that the behavior of `MIN_LANGUAGES` is different depending on the language detector:
+- The Optimaize language detector will often produce the exact number of `MIN_LANGUAGES` requested by a user.
+- The OpenNLP language detector tends to only produce 1 language unless `MIN_LANGUAGES` is set to 2 or greater. However, low confidence results are still sometimes thrown out even if the user requested additional language predictions. (This indicates that OpenNLP uses a different filtering mechanism which could be investigated further).
+
+Language results are stored as follows:
+- `TEXT_LANGUAGE` : The primary detected language for a given text. Set to "Unknown" if no language is identified.
+- `TEXT_LANGUAGE_CONFIDENCE` : A confidence setting for the primary language ranging from `NONE` to `HIGH` confidence. See [note here.](https://tika.apache.org/1.21/api/org/apache/tika/language/detect/LanguageConfidence.html)
+
+Secondary languages and their confidence scores are listed as comma separated strings:
+- `SECONDARY_TEXT_LANGUAGES` : A list of secondary languages (from greatest to least confidence) separated by ", " delimiters.
+    Set to "Unknown" if no language is identified.
+- `SECONDARY_TEXT_LANGUAGE_CONFIDENCES` : A confidence list corresponding to the secondary detected languages in order (also separated by commas). 
+
 # Supported Language Detectors:
 
 This component supports the following language detectors. Users can select their preferred detector using
--- a/java/TikaTextDetection/README.md
+++ b/java/TikaTextDetection/README.md
@@ -56,7 +56,7 @@ Please note that the behavior of `MIN_LANGUAGES` is different depending on the l
 
 Language results are stored as follows:
 - `TEXT_LANGUAGE` : The primary detected language for a given text. Set to "Unknown" if no language is identified.
-- `TEXT_LANGUAGE_CONFIDENCE` : Raw confidence score for the primary language. See [note here.](https://tika.apache.org/2.4.1/api/index.html?org/apache/tika/language/detect/LanguageResult.html)
+- `TEXT_LANGUAGE_CONFIDENCE` : Raw confidence score for the primary language. Ranges from 0 to 1 for OpenNLP and Optimaize language detectors, and set to -1 if no results are found. See [note here.](https://tika.apache.org/2.4.1/api/index.html?org/apache/tika/language/detect/LanguageResult.html)
 - `ISO_LANGUAGE` : The primary detected language for a given text in ISO 639-3 format. Set to "UNKNOWN" if no language id identified (as "UNK" is an ISO code).
 
 Secondary languages and their confidence scores are listed as comma separated strings:
@@ -87,7 +87,7 @@ the `LANG_DETECTOR` option:
   Third party language detection project that supports 71 languages.
   Predicts target language using N-gram frequency matching between input and language profiles.
   Supports almost every language present in Tika's Language Detector except Esperanto.
- Please note that Optimaize supports Punjabi/Panjabi while OpenNLP supports Western Punjabi/Panjabi.
+  Please note that Optimaize supports Punjabi/Panjabi while OpenNLP supports Western Punjabi/Panjabi.
  
    
 # Supported Language List:
--- a/java/TikaTextDetection/README.md
+++ b/java/TikaTextDetection/README.md
@@ -67,6 +67,8 @@ Secondary languages and their confidence scores are listed as comma separated st
     Set to an empty string if no language is identified.
 - `SECONDARY_TEXT_LANGUAGE_CONFIDENCES` : A confidence list corresponding to the secondary detected languages in order (also separated by commas). 
 
+  For secondary language predictions, ensure that `MIN_LANGUAGES` property is set to 2 or greater. Secondary language results are typically ignored by the component as Tika has a high confidence threshold for reasonable predictions.
+
 # Supported Language Detectors:
 
 This component supports the following language detectors. Users can select their preferred detector using
@@ -79,25 +81,18 @@ the `LANG_DETECTOR` option:
     - [Leipzig corpus](https://wortschatz.uni-leipzig.de/en/download)                                                                                                     
     - [cc-100](https://data.statmt.org/cc-100/)                                                                                                                           
                                                                                                                                                                           
-  Supports almost every language in Optimaize and Tika Language Detectors except Aragonese.    
+  Supports almost every language in Optimaize and Tika Language Detectors except Aragonese.  
 
+  **PLEASE NOTE**, if `opennlp` is selected, ensure that the `MIN_LANGUAGES` property is set to 1 or greater.
 
 - `LANG_DETECTOR = optimaize`: [Optimaize Language Detector](https://github.com/optimaize/language-detector)
 
   Third party language detection project that supports 71 languages.
   Predicts target language using N-gram frequency matching between input and language profiles.
   Supports almost every language present in Tika's Language Detector except Esperanto.
- Please note that Optimaize supports Punjabi/Panjabi while OpenNLP supports Western Punjabi/Panjabi.      
+ Please note that Optimaize supports Punjabi/Panjabi while OpenNLP supports Western Punjabi/Panjabi.
+ 
    
-                                                                                                                                                          
-- (Depreciated) `LANG_DETECTOR = tika`: [Apache Tika Language Detector](https://tika.apache.org/2.4.1/api/org/apache/tika/langdetect/tika/TikaLanguageDetector.html)     
-                                                                                                                                                           
-  Apache Tika's original in-house language detection capability.                                                                                           
-  Predicts target language using vector distance of trigrams between input string and language models.                                                     
-  Supports 28 languages (listed in following section).                                                                                                                                                                                                                                          
-  **NOTE: Developers have warned that this legacy detector is depreciated and won't work well on short snippets of text.**                                 
-
-       
 # Supported Language List:
 
 - Tika Language Detector:
--- a/java/TikaTextDetection/README.md
+++ b/java/TikaTextDetection/README.md
@@ -44,12 +44,9 @@ The following format-specific behaviors were observed using Tika 1.28.1 on Ubunt
 
 Tika supports the following language detection properties:
 
-- `MAX_REASONABLE_LANGUAGES`: Specifies maximum number of top detected languages. 
-When set to 0 or below, allow any number of language results that are marked as reasonably certain by Tika.
-
 - `MIN_LANGUAGES`: When set to a positive integer, attempt to return specified number of top languages, even if some are not marked as reasonably certain. Non-positive values disable this property to only accept reasonable predictions.
 
-For instance, if `MAX_REASONABLE_LANGUAGES` is set to 5 and `MIN_LANGUAGES` is set to 2, the component will always attempt to return the top 2 predicted languages, followed by the next 3 if they are marked as reasonably certain.
+For instance, if `MIN_LANGUAGES` is set to 2, the component will always attempt to return the top 2 predicted languages, followed by the next languages if they are also marked as reasonably certain.
 
 If `MAX_REASONABLE_LANGUAGES` is set to -1 and `MIN_LANGUAGES` is set to 2 (default), the component will always attempt to return the top predicted language and a secondary language, even if they are not set to reasonably confident by Tika.
 
@@ -59,12 +56,12 @@ Please note that the behavior of `MIN_LANGUAGES` is different depending on the l
 
 Language results are stored as follows:
 - `TEXT_LANGUAGE` : The primary detected language for a given text. Set to "Unknown" if no language is identified.
-- `TEXT_LANGUAGE_CONFIDENCE` : A confidence setting for the primary language ranging from `NONE` to `HIGH` confidence. See [note here.](https://tika.apache.org/1.21/api/org/apache/tika/language/detect/LanguageConfidence.html)
+- `TEXT_LANGUAGE_CONFIDENCE` : Raw confidence score for the primary language. See [note here.](https://tika.apache.org/2.4.1/api/index.html?org/apache/tika/language/detect/LanguageResult.html)
 - `ISO_LANGUAGE` : The primary detected language for a given text in ISO 639-3 format. Set to "UNKNOWN" if no language id identified (as "UNK" is an ISO code).
 
 Secondary languages and their confidence scores are listed as comma separated strings:
 - `SECONDARY_TEXT_LANGUAGES` : A list of secondary languages (from greatest to least confidence) separated by ", " delimiters.
-    Set to an empty string if no language is identified.
+    Note,`SECONDARY_TEXT` properties are not included if no secondary languages are identified.
 - `SECONDARY_TEXT_LANGUAGE_CONFIDENCES` : A confidence list corresponding to the secondary detected languages in order (also separated by commas). 
 
   For secondary language predictions, ensure that `MIN_LANGUAGES` property is set to 2 or greater. Secondary language results are typically ignored by the component as Tika has a high confidence threshold for reasonable predictions.
--- a/java/TikaTextDetection/README.md
+++ b/java/TikaTextDetection/README.md
@@ -39,3 +39,166 @@ The following format-specific behaviors were observed using Tika 1.28.1 on Ubunt
 
 - OpenDocument Spreadsheet documents will generate one track per cell, as well as some additional tracks with
   date and time information, "Page /", and "???".
+
+# Supported Language Detectors:
+
+This component supports the following language detectors. Users can select their preferred detector using
+the `LANG_DETECTOR` option:
+                                                                                                                                                                          
+- `LANG_DETECTOR = opennlp`: [Apache Tika OpenNLP Language Detector](https://tika.apache.org/2.4.1/api/org/apache/tika/langdetect/opennlp/OpenNLPDetector.html)           
+                                                                                                                                                                          
+  Apache Tika's latest in-house language detection capability based on OpenNLP's language detector.                                                                       
+  Uses Machine Learning (ML) models trained on the following datasets and supports 148 languages in total                                                                   
+    - [Leipzig corpus](https://wortschatz.uni-leipzig.de/en/download)                                                                                                     
+    - [cc-100](https://data.statmt.org/cc-100/)                                                                                                                           
+                                                                                                                                                                          
+  Supports almost every language in Optimaize and Tika Language Detectors except Aragonese.    
+
+
+- `LANG_DETECTOR = optimaize`: [Optimaize Language Detector](https://github.com/optimaize/language-detector)
+
+  Third party language detection project that supports 71 languages.
+  Predicts target language using N-gram frequency matching between input and language profiles.
+  Supports almost every language present in Tika's Language Detector except Esperanto.
+ Please note that Optimaize supports Punjabi/Panjabi while OpenNLP supports Western Punjabi/Panjabi.      
+   
+                                                                                                                                                          
+- (Depreciated) `LANG_DETECTOR = tika`: [Apache Tika Language Detector](https://tika.apache.org/2.4.1/api/org/apache/tika/langdetect/tika/TikaLanguageDetector.html)     
+                                                                                                                                                           
+  Apache Tika's original in-house language detection capability.                                                                                           
+  Predicts target language using vector distance of trigrams between input string and language models.                                                     
+  Supports 28 languages (listed in following section).                                                                                                                                                                                                                                          
+  **NOTE: Developers have warned that this legacy detector is depreciated and won't work well on short snippets of text.**                                 
+
+       
+# Supported Language List:
+
+- Tika Language Detector:
+  - Belarusian
+  - Catalan
+  - Danish
+  - German
+  - Esperanto
+  - Estonian
+  - Greek
+  - English
+  - Spanish
+  - Finnish
+  - French
+  - Persian
+  - Galician
+  - Hungarian
+  - Icelandic
+  - Italian
+  - Lithuanian
+  - Dutch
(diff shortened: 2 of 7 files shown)


jrobble (changes requested):
So far, I've only done a static code review.

> Reviewed 2 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/305#-:-NBn4rui3-k833IwP4c_:b-xu8m3)* status: 2 of 6 files reviewed, 15 unresolved discussions (waiting on @hhuangMITRE)

___
*[`java/TikaTextDetection/README.md` line 43 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmu08s6agE0hT8BZpa:-NBmu08s6agE0hT8BZpb:b-d2kzdx) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L43)):*
> ```Markdown
>   date and time information, "Page /", and "???".
> 
> # Language detection parameters
> ```

Capitalize every important word in the header.
___
*[`java/TikaTextDetection/README.md` line 61 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmwNZL0BA8TrdnD_VZ:-NBmwNZL0BA8TrdnD_V_:b-vpnbsg) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L61)):*
> ```Markdown
> 
> Language results are stored as follows:
> - `TEXT_LANGUAGE` : The primary detected language for a given text. Set to "Unknown" if no language is identified.
> ```

We need to ensure that we can use this output in Tika Text --> Translation pipelines. The custom translation component ( @cdglasz ) currently expects the language property to be `DECODED_LANGUAGE` or `LANGUAGE`. Also, the output is an ISO 639-3 code. For example:

```
"detectionProperties": {
   "DECODED_LANGUAGE": "SPA",
   "DECODED_LANGUAGE_CONFIDENCE": "0.9994221715043562",
```

Looking at the Tesseract component, I noticed that it outputs `TEXT_LANGUAGE` too. However, that is really a script (in all cases where OSD is successful and we're not defaulting to something - correct me if I'm wrong), for example, `script/Cyrillic`, so it's not suitable as a translation input. In a perfect world, I would rename that to `TEXT_SCRIPT`, but we can't do that right now or risk breaking compatibility with existing job parsers.

I think we should update the custom translation component to also accept `TEXT_LANGUAGE` as an input. If someone tries to create a Tesseract --> Translation pipeline using the custom translation component that will cause an error saying the language (really a script), is not supported. I think that's fine. 
___
*[`java/TikaTextDetection/README.md` line 65 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmvk2f7qf19iswKPok:-NBmltHlEpA3qzuujzyz:b-mmc78y) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L65)):*
> ```Markdown
> 
> Secondary languages and their confidence scores are listed as comma separated strings:
> - `SECONDARY_TEXT_LANGUAGES` : A list of secondary languages (from greatest to least confidence) separated by ", " delimiters.
> ```

I just want to point out that our use of semi-colons for `PAGE_NUM` is inconsistent with our use of commas for this output and the associated confidences, but don't change it.

Our custom speech-to-text components also use commas for the detected languages and associated confidences. It's best to be consistent with that for this language detection capability. Also, I don't want to change `PAGE_NUM` to use commas now since it might break compatibility with existing job parsers.
___
*[`java/TikaTextDetection/README.md` line 135 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmruQmD51fKMEWHbw1:-NBmruQmD51fKMEWHbw2:bdaumzc) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L135)):*
> ```Markdown
> - Optimaize Language Detector:
>   - Every language in Tika's language detector, except Esperanto.
>   - Afrikaans
> ```

Listing these in alphabetical order will help readability. 
___
*[`java/TikaTextDetection/README.md` line 185 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrzKU4rrsrO741syD:-NBmrzKU4rrsrO741syE:bdaumzc) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L185)):*
> ```Markdown
> - OpenNLP Language Detector:
>   - Every language in Tika and Optimaize Language Detectors except Aragonese. 
>   - Bihari languages 
> ```

Listing these in alphabetical order will help readability. 
___
*[`java/TikaTextDetection/plugin-files/descriptor/descriptor.json` line 35 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmm6OS3Y8UegnuDOh0:-NBmm6OS3Y8UegnuDOh1:b98d34y) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/plugin-files/descriptor/descriptor.json#L35)):*
> ```JSON
>         },
>         {
>           "name": "LANG_DETECTOR",
> ```

Use `LANGUAGE_DETECTOR" for consistency with other parameters.
___
*[`java/TikaTextDetection/plugin-files/descriptor/descriptor.json` line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmmFs88_zfszdm324m:-NBmmFs88_zfszdm324n:b-q3qne) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/plugin-files/descriptor/descriptor.json#L36)):*
> ```JSON
>         {
>           "name": "LANG_DETECTOR",
>           "description": "Specifies which Tika Language detector to use. Current options are `opennlp`, `optimaize` and `tika` (Note: `tika` is depreciated and does not work for short text). Defaults to `opennlp`.",
> ```

Use lowercase L in "Language detector".

Don't need to mention default in the description. 
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 140 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmnl7rD_kEDIV9RkZL:-NBmnl7rD_kEDIV9RkZM:b-ksujs4) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L140)):*
> ```Java
>         if (!supportedLangDetectors.contains(langOption))
>         {
>             log.warn(langOption + " language option not supported. Defaulting to `opennlp` language detector.");
> ```

Say "language detection option".
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 146 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmoF7C8EE7YxbeJMhF:-NBmoF7C8EE7YxbeJMhG:bbmbllg) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L146)):*
> ```Java
>             identifier = new OptimaizeLangDetector();
>             try {
>                 identifier.loadModels();
> ```

How long does this call take? Is it something we want to cache between job runs?

More generally, is the initialization time for **any** of the language detectors something that we want to do only once and only load a new detector if different from the last job run? Caching is not necessary if the overhead is trivial.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 154 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmo-ct2BVZUV5tZF7W:-NBmo-ct2BVZUV5tZF7X:b6hn01f) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L154)):*
> ```Java
>         }
>         else if(langOption == "tika") {
>             //TODO: It appears that the legaacy tika detector has several issues procssing short sections of text.
> ```

Spelling: "procssing"
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 207 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmpAQh2518t1kyhL2p:-NBmpAQh2518t1kyhL2q:bc0rti2) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L207)):*
> ```Java
>                 if (text.length() >= charLimit) {
>                     List<LanguageResult> langResultList;
>                     langResultList = identifier.detectAll(text);
> ```

This line can be combined with the one above it.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 212 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmqEPmEXErQNFoSxZh:-NBmqEPmEXErQNFoSxZi:benh6gx) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L212)):*
> ```Java
>                     List<String> confidenceList = new ArrayList<String>();
> 
>                     maxLang = maxLang > langResultList.size() || maxLang < 1 ? langResultList.size(): maxLang;
> ```

Please define new variables for `maxLang` and `minLang` here so we don't overwrite what we read in as parameters. I've encountered situations where developers assume input parameters are never reset, which has led to bugs in the code.

Also, parenthesis around the conditional part of the ternary operator (before the ?) will help readability.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 238 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBn3MvI7W7srxaN_hQm:-NBn3MvI7W7srxaN_hQn:bz2ue53) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L238)):*
> ```Java
> 
>                     if(langList.size() < 1) {
>                         genericDetectionProperties.put("TEXT_LANGUAGE", "Unknown");
> ```

Returning `Unknown` here could mess up the custom translator. We should discuss with @cdglasz . We may want the custom translator to ignore that rather than treat it as an actual language. I think some of the custom speech-to-text components may produce some form of unknown as the decoded language.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 241 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrXLJ0l0UH8FW4Opj:-NBmrXLJ0l0UH8FW4Opk:bw0gmby) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L241)):*
> ```Java
>                         genericDetectionProperties.put("TEXT_LANGUAGE_CONFIDENCE", "-1");
>                         genericDetectionProperties.put("SECONDARY_TEXT_LANGUAGES", "Unknown");
>                         genericDetectionProperties.put("SECONDARY_TEXT_LANGUAGE_CONFIDENCES", "NONE");
> ```

`NONE` seems odd since a list of numerical values is expected. The Tesseract component just omits the following detection properties if they're not suitable:

```
        if (secondary_scripts.size() > 0) {
            // ...
            detection_properties["OSD_SECONDARY_SCRIPTS"] = boost::algorithm::join(scripts, ", ");
            detection_properties["OSD_SECONDARY_SCRIPT_SCORES"] = boost::algorithm::join(scores, ", ");
```
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 283 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmmoUWDeLwvBGZ0--r:-NBmmoUWDeLwvBGZ0--s:b9acikz) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L283)):*
> ```Java
> 
>     // Map for translating from ISO 639-2 code to english description.
>     private static Map<String,String> convertISO639() {
> ```

This method prepares a map for use in a later conversion (vs. performing a conversation itself at the moment it's called), so I prefer the old method name of `initLangMap`.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 358 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrehC1y5iTIX413MY:-NBmrehC1y5iTIX413MZ:b-xwq6vl) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L358)):*
> ```Java
>         map.put("zh-tw", "Traditional Chinese");
> 
>         map.put("bih", "Bihari languages");
> ```

Putting these in alphabetical order will help readability.

jrobble:
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 150 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBrZpSB52VzEcfRgY-S:-NBrZpSB52VzEcfRgY-T:b-sh3t0b) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L150)):*
> ```Java
>                 String errorMsg = "Error loading Optimaize Language Model.";
>                 log.error(errorMsg, e);
>                 throw new MPFComponentDetectionError(MPFDetectionError.MPF_OTHER_DETECTION_ERROR_TYPE, errorMsg);
> ```

Please use `MPF_DETECTION_NOT_INITIALIZED` instead of `MPF_OTHER_DETECTION_ERROR_TYPE` when failing to load the models.

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/305)* status: 2 of 6 files reviewed, 16 unresolved discussions (waiting on @cdglasz, @hhuangMITRE, and @jrobble)

___
*[`java/TikaTextDetection/README.md` line 43 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmu08s6agE0hT8BZpa:-NBtpiqo9IYUWDB2wFHZ:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L43)):*

Capitalize every important word in the header.
</blockquote></details>

Done.
___
*[`java/TikaTextDetection/README.md` line 61 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmwNZL0BA8TrdnD_VZ:-NBtpkTJ7l62kLGidNNP:b-eagu27) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L61)):*

We need to ensure that we can use this output in Tika Text --> Translation pipelines. The custom translation component ( @cdglasz ) currently expects the language property to be `DECODED_LANGUAGE` or `LANGUAGE`. Also, the output is an ISO 639-3 code. For example:

```
"detectionProperties": {
   "DECODED_LANGUAGE": "SPA",
   "DECODED_LANGUAGE_CONFIDENCE": "0.9994221715043562",
```

Looking at the Tesseract component, I noticed that it outputs `TEXT_LANGUAGE` too. However, that is really a script (in all cases where OSD is successful and we're not defaulting to something - correct me if I'm wrong), for example, `script/Cyrillic`, so it's not suitable as a translation input. In a perfect world, I would rename that to `TEXT_SCRIPT`, but we can't do that right now or risk breaking compatibility with existing job parsers.

I think we should update the custom translation component to also accept `TEXT_LANGUAGE` as an input. If someone tries to create a Tesseract --> Translation pipeline using the custom translation component that will cause an error saying the language (really a script), is not supported. I think that's fine. 
</blockquote></details>

@jrobble Yes, the Tesseract component technically detects scripts (rather than exact language) using the OSD capability. 

As a follow up: 

Discussed with Chris our options for a compatible text language field.

Currently, we've settled on `ISO_LANGUAGE` for Tika with missing language detections set to `UNKNOWN`. 

We're going to continue passing over `TEXT_LANGUAGE` from Tesseract, with the assumption that the component will sometimes output ISO formatted languages (when specified by user). 
___
*[`java/TikaTextDetection/README.md` line 135 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmruQmD51fKMEWHbw1:-NBtrXvg1PzdgXg35EoX:b-e7fxuc) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L135)):*

Listing these in alphabetical order will help readability. 
</blockquote></details>

Resorted!
___
*[`java/TikaTextDetection/README.md` line 185 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrzKU4rrsrO741syD:-NBtr_bW4ybrTNBsciUx:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/README.md#L185)):*

Listing these in alphabetical order will help readability. 
</blockquote></details>

Done.
___
*[`java/TikaTextDetection/plugin-files/descriptor/descriptor.json` line 35 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmm6OS3Y8UegnuDOh0:-NBtrb879E6NrGhla3aO:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/plugin-files/descriptor/descriptor.json#L35)):*

Use `LANGUAGE_DETECTOR" for consistency with other parameters.
</blockquote></details>

Done.
___
*[`java/TikaTextDetection/plugin-files/descriptor/descriptor.json` line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmmFs88_zfszdm324m:-NBtrcMi8b-1zgpSeXye:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/plugin-files/descriptor/descriptor.json#L36)):*

Use lowercase L in "Language detector".

Don't need to mention default in the description. 
</blockquote></details>

Done.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 140 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmnl7rD_kEDIV9RkZL:-NBtrdJq4Eve1o-E2ESu:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L140)):*

Say "language detection option".
</blockquote></details>

Done.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 146 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmoF7C8EE7YxbeJMhF:-NBtcADP68MrtI7PGHb3:b-eh5hqw) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L146)):*

How long does this call take? Is it something we want to cache between job runs?

More generally, is the initialization time for **any** of the language detectors something that we want to do only once and only load a new detector if different from the last job run? Caching is not necessary if the overhead is trivial.
</blockquote></details>

I've done a quick timing test, around 2-3 seconds for opennlp and 0.5 seconds for optimaize.

When multiple job calls are run in succession, the initialization time drops down below 1 millisecond. 

Tika has reported that opennlp is initialized with a static call, and since we only need the detector to be initialized once, I've updated the component to handle the language detector calls once when the component is initialized. 
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 150 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBrZpSB52VzEcfRgY-S:-NBtrj0y-NiVNuDicVWk:bei6zrn) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L150)):*

Please use `MPF_DETECTION_NOT_INITIALIZED` instead of `MPF_OTHER_DETECTION_ERROR_TYPE` when failing to load the models.
</blockquote></details>

Done. Relocated to component initalization.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 154 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmo-ct2BVZUV5tZF7W:-NBtroET1MGA1pmpH4Jf:b-hkjhw2) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L154)):*

Spelling: "procssing"
</blockquote></details>

Whoops, thanks for catching that. I spotted another typo here and fixed that as well.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 207 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmpAQh2518t1kyhL2p:-NBtsT2i9FU_YlNXOgqd:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L207)):*

This line can be combined with the one above it.
</blockquote></details>

Done.



_Code quote:_
```Java
langResultList = identifier.detectAll(text);
```
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 212 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmqEPmEXErQNFoSxZh:-NBtsYs87maWH-hoS_KF:b-e5bud5) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L212)):*

Please define new variables for `maxLang` and `minLang` here so we don't overwrite what we read in as parameters. I've encountered situations where developers assume input parameters are never reset, which has led to bugs in the code.

Also, parenthesis around the conditional part of the ternary operator (before the ?) will help readability.
</blockquote></details>

Thanks for the feedback. I agree that we'd want to avoid that confusion and it is good coding practice in general. 
I've also updated the ternary operator code for clarity.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 238 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBn3MvI7W7srxaN_hQm:-NBtt2G0AM3ozHEHzuNZ:btp0n80) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L238)):*

Returning `Unknown` here could mess up the custom translator. We should discuss with @cdglasz . We may want the custom translator to ignore that rather than treat it as an actual language. I think some of the custom speech-to-text components may produce some form of unknown as the decoded language.
</blockquote></details>

Update: Synched with Chris and identified that we're using `UNKNOWN` keyword for blank language detections. Set this component return that in the `ISO_LANGUAGE` field.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 241 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrXLJ0l0UH8FW4Opj:-NBttKdq1NtPT4a7k6af:b617knl) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L241)):*

`NONE` seems odd since a list of numerical values is expected. The Tesseract component just omits the following detection properties if they're not suitable:

```
        if (secondary_scripts.size() > 0) {
            // ...
            detection_properties["OSD_SECONDARY_SCRIPTS"] = boost::algorithm::join(scripts, ", ");
            detection_properties["OSD_SECONDARY_SCRIPT_SCORES"] = boost::algorithm::join(scores, ", ");
```
</blockquote></details>

I agree it's a bit strange, I can look into seeing if the language detectors offer more info. Right now only Optimiaze appears to be offering confidence results, but they've been sorted into the following labels by Tika:

`"NONE, LOW, MEDIUM, HIGH"`

https://tika.apache.org/1.21/api/org/apache/tika/language/detect/LanguageConfidence.html

If non-normalized confidence values are encountered, perhaps we should just leave this field out. I'll also check if OpenNLP has confidence parameters to be returned as well in a follow-up update.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 283 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmmoUWDeLwvBGZ0--r:-NBttuuCBzzy4AgOLyzQ:b-ujc5ai) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L283)):*

This method prepares a map for use in a later conversion (vs. performing a conversation itself at the moment it's called), so I prefer the old method name of `initLangMap`.
</blockquote></details>

Done. Also added the isoMap as well for mapping between ISO 639-1 and ISO 639-3. 
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 358 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrehC1y5iTIX413MY:-NBtu9E39l1OIdDKKw7J:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L358)):*

Putting these in alphabetical order will help readability.
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/305)* status: 4 of 6 files reviewed, 7 unresolved discussions (waiting on @cdglasz and @hhuangMITRE)

___
*[`java/TikaTextDetection/plugin-files/descriptor/descriptor.json` line 35 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6e2t27u0GzZjY9FVE:-NC6e2t3-_05b6kIkxVI:b-hduosb) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/plugin-files/descriptor/descriptor.json#L35)):*
> ```JSON
>         },
>         {
>           "name": "LANGUAGE_DETECTOR",
> ```

Should we just remove `tika` support if it's inferior and depreciated? As long as the component output for openNLP and optimaize is similar, we don't need to worry about breaking compatibility.

If we want to keep it, we need a unit test for it.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 143 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6fyTk43vu9_rMMI7h:-NC6fyTl3xkZBvcBmma4:bhe9js) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L143)):*
> ```Java
>         }
>         LanguageDetector identifier;
>         if (langOption == "optimaize") {
> ```

You need to use `.equals()` for string comparison in Java.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 146 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmoF7C8EE7YxbeJMhF:-NC6cEo8DOLp5DsOoZxy:b-us7klt) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L146)):*

I've done a quick timing test, around 2-3 seconds for opennlp and 0.5 seconds for optimaize.

When multiple job calls are run in succession, the initialization time drops down below 1 millisecond. 

Tika has reported that opennlp is initialized with a static call, and since we only need the detector to be initialized once, I've updated the component to handle the language detector calls once when the component is initialized. 


</blockquote></details>

Another overhead consideration is the amount of memory loading each of these takes. I ran a profiler. Note the highlighted line and the one below:



OpenNLP requires ~318 MiB
Optimaize requires ~263 MiB

That's a lot of memory for something that might not even be used, depending on the job properties. Please only load the required detector based on job properties, and switch to loading the other detector if the next job has a different value for that property.

We can't force Java to unload the other detector (easily), but if you have one `LanguageDetector langDetector` class variable in `TikaTextDetectionComponent` and overwrite it with the required detector, the previously loaded detector will go out of scope and be garbage collected eventually.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 153 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6g3-X3EFiWpqsz6EZ:-NC6g3-X3EFiWpqsz6E_:bhe9js) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L153)):*
> ```Java
>             }
>         }
>         else if(langOption == "tika") {
> ```

You need to use `.equals()` for string comparison in Java.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 241 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrXLJ0l0UH8FW4Opj:-NC66eLR2S-HBjd7lr_Y:bhkr00h) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L241)):*

I agree it's a bit strange, I can look into seeing if the language detectors offer more info. Right now only Optimiaze appears to be offering confidence results, but they've been sorted into the following labels by Tika:

`"NONE, LOW, MEDIUM, HIGH"`

https://tika.apache.org/1.21/api/org/apache/tika/language/detect/LanguageConfidence.html

If non-normalized confidence values are encountered, perhaps we should just leave this field out. I'll also check if OpenNLP has confidence parameters to be returned as well in a follow-up update.
</blockquote></details>

The `LanguageResult` class has a `getRawScore()` method, which appears to have a valid value for openNLP and optimaize between 0 and 1 (looking at the decompiled .class files for how the confidence string is determined). I think it's worth returning this raw score as the detection confidence instead of the string value.

Another reason for this change is that it seems odd that when you set `MIN_LANGUAGES > 0` the associated `TEXT_LANGUAGE_CONFIDENCE` can be`NONE` for a valid `TEXT_LANGUAGE`.

---

I got openNLP to return a `LOW` confidence (as opposed to `NONE`) when providing it with a lot of Chinese text (10,368 characters). Not saying that this has anything to do with Chinese specifically, rather, that's a lot of characters for a `LOW` confidence.

Internally, it uses this calculation:

```
    private static LanguageConfidence getConfidence(double confidence) {
        if (confidence > 0.9) {
            return LanguageConfidence.HIGH;
        } else if (confidence > 0.85) {
            return LanguageConfidence.MEDIUM;
        } else {
            return confidence > 0.2 ? LanguageConfidence.LOW : LanguageConfidence.NONE;
        }
    }
```

Note that `LanguageResult.isReasonablyCertain()` is defined as:

```
    public boolean isReasonablyCertain() {
        return this.confidence == LanguageConfidence.HIGH;
    }
```

What I'm saying is, I don't think we want to rely on using `isReasonablyCertain()` for openNLP. It's too selective. I don't know how many characters it would take to achieve a `HIGH` confidence. I assume a lot.

Not saying this issue requires a code change. Just saying that we always want to make sure `MIN_LANGUAGES` is 1 or greater when OpenNLP is selected. We should probably document that.

---

If we don't set a cap via `MAX_REASONABLE_LANGUAGES`, how many languages have you ever seen it return as reasonable? I wonder if we can simplify things and not cap this. Then we wouldn't need the property.
___
*[`java/TikaTextDetection/src/test/java/org/mitre/mpf/detection/tika/TestTikaTextDetectionComponent.java` line 159 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6XCgT9hz_iotBDAzi:-NC6XCgT9hz_iotBDAzj:b-8kty7y) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/src/test/java/org/mitre/mpf/detection/tika/TestTikaTextDetectionComponent.java#L159)):*
> ```Java
> 
>         MPFGenericJob genericJob = new MPFGenericJob("TestGenericJob", mediaPath, jobProperties, mediaProperties);
>         boolean debug = false;
> ```

Remove this debug line.

jrobble (changes requested):
> Reviewed 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/305)* status: all files reviewed, 9 unresolved discussions (waiting on @cdglasz and @hhuangMITRE)

___
*[`java/TikaTextDetection/README.md` line 66 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC65g9K2oQ2qVhaxxDD:-NC65g9K2oQ2qVhaxxDE:b-sctohx) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/README.md#L66)):*
> ```Markdown
> 
> Secondary languages and their confidence scores are listed as comma separated strings:
> - `SECONDARY_TEXT_LANGUAGES` : A list of secondary languages (from greatest to least confidence) separated by ", " delimiters.
> ```

 Just omit `SECONDARY_TEXT_LANGUAGES` and `SECONDARY_TEXT_LANGUAGE_CONFIDENCES` if they are not used , as Tesseract does for `OSD_SECONDARY_SCRIPTS` and`OSD_SECONDARY_SCRIPT_SCORES`. I don't think I was clear before. What I mean by "omit" is "not include them in the detection property map at all".

Also, I would feel better setting `TEXT_LANGUAGE_CONFIDENCE` to -1 for `Unknown` rather than leaving it blank.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 303 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC675EhAd-zAel86n4Q:-NC675Ei34Y-M4pVj1S0:baxh12t) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L303)):*
> ```Java
> 
>     // Map for translating from ISO 639-2 code to english description.
>     private static Map<String,String> initLangMap() {
> ```

Consider moving these mappings to a util class since they take up a lot of space.

hhuangMITRE (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/305)* status: 1 of 7 files reviewed, 9 unresolved discussions (waiting on @cdglasz, @hhuangMITRE, and @jrobble)

___
*[`java/TikaTextDetection/README.md` line 66 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC65g9K2oQ2qVhaxxDD:-NCsJzr15ML1Dt4qZcm8:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/README.md#L66)):*

 Just omit `SECONDARY_TEXT_LANGUAGES` and `SECONDARY_TEXT_LANGUAGE_CONFIDENCES` if they are not used , as Tesseract does for `OSD_SECONDARY_SCRIPTS` and`OSD_SECONDARY_SCRIPT_SCORES`. I don't think I was clear before. What I mean by "omit" is "not include them in the detection property map at all".

Also, I would feel better setting `TEXT_LANGUAGE_CONFIDENCE` to -1 for `Unknown` rather than leaving it blank.
</blockquote></details>

Done.
___
*[`java/TikaTextDetection/plugin-files/descriptor/descriptor.json` line 35 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6e2t27u0GzZjY9FVE:-NCT5BX1EeXI2jCtvBCA:blklz3g) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/plugin-files/descriptor/descriptor.json#L35)):*

Should we just remove `tika` support if it's inferior and depreciated? As long as the component output for openNLP and optimaize is similar, we don't need to worry about breaking compatibility.

If we want to keep it, we need a unit test for it.
</blockquote></details>

I agree with removing it, documentation (https://tika.apache.org/2.4.1/api/org/apache/tika/langdetect/tika/TikaLanguageDetector.html) lists it as obsolete for the time being. 

We can revisit `tika` if Apache decides to update it, but it seems that they've switched over to OpenNLP for the latest language/text processing capabilities. 
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 143 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6fyTk43vu9_rMMI7h:-NCT6LwP5y-mkbr3mwKY:bj6veeo) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L143)):*

You need to use `.equals()` for string comparison in Java.
</blockquote></details>

Thanks for the reminder, I've fixed this and the line below.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 146 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmoF7C8EE7YxbeJMhF:-NCT66sT3AjGLwv7hBD_:by5zmtd) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L146)):*

Another overhead consideration is the amount of memory loading each of these takes. I ran a profiler. Note the highlighted line and the one below:



OpenNLP requires ~318 MiB
Optimaize requires ~263 MiB

That's a lot of memory for something that might not even be used, depending on the job properties. Please only load the required detector based on job properties, and switch to loading the other detector if the next job has a different value for that property.

We can't force Java to unload the other detector (easily), but if you have one `LanguageDetector langDetector` class variable in `TikaTextDetectionComponent` and overwrite it with the required detector, the previously loaded detector will go out of scope and be garbage collected eventually.
</blockquote></details>

Thanks for checking the memory usage. I agree with switching between detectors rather than caching both and have updated the code.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 153 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6g3-X3EFiWpqsz6EZ:-NCTEQvy4ybDgy7zWF2D:bj3eodx) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L153)):*

You need to use `.equals()` for string comparison in Java.
</blockquote></details>

Fixed, thanks!
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 241 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NBmrXLJ0l0UH8FW4Opj:-NCTHk1X6a3HgNnFoBZt:bccfjjo) ([raw file](https://github.com/openmpf/openmpf-components/blob/cfba71141f48c2734594a6385ebf9596b9af2663/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L241)):*

The `LanguageResult` class has a `getRawScore()` method, which appears to have a valid value for openNLP and optimaize between 0 and 1 (looking at the decompiled .class files for how the confidence string is determined). I think it's worth returning this raw score as the detection confidence instead of the string value.

Another reason for this change is that it seems odd that when you set `MIN_LANGUAGES > 0` the associated `TEXT_LANGUAGE_CONFIDENCE` can be`NONE` for a valid `TEXT_LANGUAGE`.

---

I got openNLP to return a `LOW` confidence (as opposed to `NONE`) when providing it with a lot of Chinese text (10,368 characters). Not saying that this has anything to do with Chinese specifically, rather, that's a lot of characters for a `LOW` confidence.

Internally, it uses this calculation:

```
    private static LanguageConfidence getConfidence(double confidence) {
        if (confidence > 0.9) {
            return LanguageConfidence.HIGH;
        } else if (confidence > 0.85) {
            return LanguageConfidence.MEDIUM;
        } else {
            return confidence > 0.2 ? LanguageConfidence.LOW : LanguageConfidence.NONE;
        }
    }
```

Note that `LanguageResult.isReasonablyCertain()` is defined as:

```
    public boolean isReasonablyCertain() {
        return this.confidence == LanguageConfidence.HIGH;
    }
```

What I'm saying is, I don't think we want to rely on using `isReasonablyCertain()` for openNLP. It's too selective. I don't know how many characters it would take to achieve a `HIGH` confidence. I assume a lot.

Not saying this issue requires a code change. Just saying that we always want to make sure `MIN_LANGUAGES` is 1 or greater when OpenNLP is selected. We should probably document that.

---

If we don't set a cap via `MAX_REASONABLE_LANGUAGES`, how many languages have you ever seen it return as reasonable? I wonder if we can simplify things and not cap this. Then we wouldn't need the property.
</blockquote></details>

For most cases, I've also only seen the top language is marked as reasonable, and sometimes even that is not guaranteed if the text is short (in several cases, both Optimaize and OpenNLP still have the correct prediction). 

I think Tika has set a high threshold in order to ensure that all reasonable predictions are as advertised. 
As a result, I agree with leaving reasonably certain languages uncapped.

Also, double checked the `getRawScore()` method for both language detectors, and switched the language confidence properties to this option.
___
*[`java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java` line 303 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC675EhAd-zAel86n4Q:-NCTE5IH-3TSGSpvMBku:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/src/main/java/org/mitre/mpf/detection/tika/TikaTextDetectionComponent.java#L303)):*

Consider moving these mappings to a util class since they take up a lot of space.
</blockquote></details>

Done.
___
*[`java/TikaTextDetection/src/test/java/org/mitre/mpf/detection/tika/TestTikaTextDetectionComponent.java` line 159 at r2](https://reviewable.io/reviews/openmpf/openmpf-components/305#-NC6XCgT9hz_iotBDAzi:-NCTE6-C28dz5r-RnZ-Z:b-896fix) ([raw file](https://github.com/openmpf/openmpf-components/blob/03bbb28dec84f8d975ee7113bd12216be018f33b/java/TikaTextDetection/src/test/java/org/mitre/mpf/detection/tika/TestTikaTextDetectionComponent.java#L159)):*

Remove this debug line.
</blockquote></details>

Done.

PR #333, opened 2023-06-27
Catch resize exception and release client lock.

jrobble:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1697

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/333)

### DIFF
--- a/cpp/OcvYoloDetection/Frame.cpp
+++ b/cpp/OcvYoloDetection/Frame.cpp
@@ -26,6 +26,8 @@
 
 #include "Frame.h"
 
+#include <MPFDetectionException.h>
+
 using namespace MPF::COMPONENT;
 
 cv::Mat Frame::getDataAsResizedFloat(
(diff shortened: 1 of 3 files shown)


brosenberg42 (changes requested):
We should be throwing `MPFDetectionException(MPF_BAD_FRAME_SIZE, "<explanation>")`. We can check the dimensions and throw before calling cv::resize. An other option is to put a try block around just the `cv::resize` call, catch `cv::Exception`, and throw the mpf exception.

> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/333#-:-NZ0eXjS98xFeqoETjHn:bqudxz9)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jrobble)

___
*[`cpp/OcvYoloDetection/triton/TritonInferencer.cpp` line 265 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/333#-NZ0gbLE6MlbgKzBuRHE:-NZ0gbLE6MlbgKzBuRHF:bupyoiw) ([raw file](https://github.com/openmpf/openmpf-components/blob/3a1227b7068fdbec895ad661ec5486e5b9965095/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L265)):*
> ```cpp
>             }
>         } 
>         catch (const std::exception &ex) {
> ```

In addition to my comments above, change this to:
```
catch (...) {
    releaseClientId(clientId, nullptr);
    throw;
}
```
We catch `...` because we never want to be deadlocked. Ideally, we would be calling `releaseClientId` in a destructor for a ClientLease object, but that is a bigger change.

`std::current_exception()` is for when you want access to an exception outside of its catch block. It is usually used to move an exception between threads.

jrobble (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/333)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @brosenberg42 and @jrobble)

___
*[`cpp/OcvYoloDetection/triton/TritonInferencer.cpp` line 265 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/333#-NZ0gbLE6MlbgKzBuRHE:-NZ1TADg71nUZslv9LGU:b-ror10o) ([raw file](https://github.com/openmpf/openmpf-components/blob/3a1227b7068fdbec895ad661ec5486e5b9965095/cpp/OcvYoloDetection/triton/TritonInferencer.cpp#L265)):*

In addition to my comments above, change this to:
```
catch (...) {
    releaseClientId(clientId, nullptr);
    throw;
}
```
We catch `...` because we never want to be deadlocked. Ideally, we would be calling `releaseClientId` in a destructor for a ClientLease object, but that is a bigger change.

`std::current_exception()` is for when you want access to an exception outside of its catch block. It is usually used to move an exception between threads.
</blockquote></details>

I did it a different way, as discussed.

jrobble (commented):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/333)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jrobble)

PR #356, opened 2024-03-23
Update Azure Translation's 50k char limit.

hhuangMITRE:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1771

So far, Azure's 50k character limit behaves as advertised - confirmed that the invisible reverse chars (i.e. `'RIGHT-TO-LEFT OVERRIDE' (U+202E)`) are also counted as part of the character length limit. 

Component tests currently indicate that such characters are properly accounted for during `len()` checks. Updating the char limit to 50k and including all characters of concern (arabic - reversed chars, invisible (U+202E), regular text) did not produce any issues and returned results from Azure. 

We can update the char limit as a baseline update for the component and add improvements for text splitting in a future `develop` PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/356)

### DIFF
--- a/python/AzureTranslation/acs_translation_component/acs_translation_component.py
+++ b/python/AzureTranslation/acs_translation_component/acs_translation_component.py
@@ -433,7 +433,7 @@ class BreakSentenceClient:
 
     # ACS limits the number of characters that can be translated in a single /translate call.
     # Taken from https://docs.microsoft.com/en-us/azure/cognitive-services/translator/reference/v3-0-translate
-    TRANSLATION_MAX_CHARS = 10_000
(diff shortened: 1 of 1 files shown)


hhuangMITRE:
Closing this PR immediately, Brian noted that we'll also need to remove Azure's Sentence Break (`_send_break_sentence_request`) logic with this change, as it would never get used with the 50k character limit triggering manual sentence break guessing.

Since that alters how the behavior functions. These changes will now get bundled into our upcoming sentence break PR for Azure, landing to `develop`.

PR #376, opened 2024-06-12
Use uppercase values for tags.

jrobble:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1822
- https://github.com/openmpf/openmpf/issues/1823

**Related PRs:**

- https://github.com/openmpf/openmpf-components/pull/376
- https://github.com/openmpf/openmpf/pull/1821

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/376)


brosenberg42 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/376)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @jrobble)

___
*[`cpp/KeywordTagging/KeywordTagging.cpp` line 294 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/376#-O-HEqp897hZjmXQ2t_R:-O-HEqp897hZjmXQ2t_S:b-f8lqru) ([raw file](https://github.com/openmpf/openmpf-components/blob/6c475d1c2420808f0a0fe91b4e5637537da0b377/cpp/KeywordTagging/KeywordTagging.cpp#L294)):*
> ```cpp
>         auto key = boost::locale::to_upper(kv.first);
>         auto values = kv.second;
>         map<wstring, vector<string>> trigger_words_offset; // sorted alphabetically by default
> ```

It isn't clear if this comment applies to keys in the map, the vectors, or both. Also, the "default" sort order is lexicographical, not alphabetical. They are equivalent in this case because they are all upper case, but I wouldn't consider that the "default" behavior since you had to do something to get it to work (calling to_upper).
___
*[`cpp/KeywordTagging/KeywordTagging.cpp` line 515 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/376#-O-HGmnP7h3JqwavUmOO:-O-HGmnP7h3JqwavUmOP:b-uxtt6a) ([raw file](https://github.com/openmpf/openmpf-components/blob/6c475d1c2420808f0a0fe91b4e5637537da0b377/cpp/KeywordTagging/KeywordTagging.cpp#L515)):*
> ```cpp
> 
>     bool has_text = false;
>     set<wstring> all_found_tags; // sorted alphabetically by default
> ```

Same issue as above. It is sorted alphabetically because you made sure all of the elements are the same case, not because of the default behavior.
___
*[`python/TransformerTagging/transformer_tagging_component/transformer_tagging_component.py` line 142 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/376#-O-HAuRw8NmkAo5_XR8d:-O-HAuRw8NmkAo5_XR8e:bysaz82) ([raw file](https://github.com/openmpf/openmpf-components/blob/6c475d1c2420808f0a0fe91b4e5637537da0b377/python/TransformerTagging/transformer_tagging_component/transformer_tagging_component.py#L142)):*
> ```Python
>             
>         for prop_to_tag in input_texts:
>             self._add_tags_for_prop(config, corpus, ff_props, prop_to_tag, input_texts[prop_to_tag])
> ```

This loop should use .items():
```python
        for prop_to_tag, input_text in input_texts.items():
            self._add_tags_for_prop(config, corpus, ff_props, prop_to_tag, input_text)
```
___
*[`python/TransformerTagging/transformer_tagging_component/transformer_tagging_component.py` line 210 at r1](https://reviewable.io/reviews/openmpf/openmpf-components/376#-O-H9uao5Nj_oyaDr8Qp:-O-H9uapDIcaxBG4mEq4:b2c642i) ([raw file](https://github.com/openmpf/openmpf-components/blob/6c475d1c2420808f0a0fe91b4e5637537da0b377/python/TransformerTagging/transformer_tagging_component/transformer_tagging_component.py#L210)):*
> ```Python
>             if "TAGS" in ff_props:
>                 # only add tag if it is not already in ff_props["TAGS"], else do nothing
>                 if tag.upper() not in ff_props["TAGS"].upper():
> ```

This won't work properly when one tag is a substring of another. We should be adding the tags to set before sorting and combining them in to a string.

jrobble (commented):
> Reviewed 5 of 9 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/376)* status: all files reviewed, 4 unresolved discussions (waiting on @brosenberg42)

brosenberg42 (approved):
> Reviewed 5 of 9 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/376)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jrobble)

PR #428, opened 2026-03-16
Use setuptools that supports pkg_resources.

jrobble:
**Issues:**

- https://github.com/openmpf/openmpf/issues/2048

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf-components/428)


jrobble:
OBE by https://github.com/openmpf/openmpf-components/pull/430

